Issue #10, opened 2015-07-23
Whitelisting doesn't work


caillette:
Nice tool. We needed that. 
Unfortunately whitelisting doesn't seem to work. Could you have a look at those test cases?
https://github.com/caillette/fast-classpath-scanner/commit/0922b9f71d1d0d79a41dc83c6e0e5c3d538fe83c

caillette:
Looking at the code it seems like whitelisting applies only to archive content, not on real directories.

lukehutch:
Ah, sorry about this. I was cleaning up the whitelisting code recently, and must have broken something. I'll try to get this fixed ASAP, although I have a couple of projects that require my time right now, so it might not be for a few days. If you already spotted how to fix it, I'd love a patch.

Also, I appreciate the test code, the lack of tests was always a glaring omission. I'll probably pull your patch into the project as a starting point. If there are other tests you plan to write, or other things you'd like to have tested, that code would be highly appreciated too!

lukehutch:
I added unit tests now and made a new release (sorry for the slight API change in 1.4.0). Thanks for showing how tests should be implemented with JUnit.

caillette:
Thanks, it works, whitelisting made my classpath scan lightning fast. 

I really like the API which preserves fast scanning but remains expressive. 

I just discovered that I should use Javassist instead of Fast-classpath-scanner because in fact I want a dependency tree. The purpose is to calculate a SHA256 fingerprint on the class implementations for financial algorithms that may evolve over time, but must always give the same result for former data. So for each version there is a top-level implementation class, which calculates its fingerprint basing on the byte array of some well-known classes and their subclasses. Alas, using well-known classes requires too much knowledge about my own code, and I'll probably mess it up one day. First I thought that dependency crawling would be a nice feature to add to Fast-classpath-scanner, but it requires full class parsing, which defeats original vision. Javassist is better suited for that. 

I'm happy to have helped with some tests, and I'll probably use Fast-classpath-scanner later for annotation scan.

lukehutch:
If you just need to know if the classes have changed, and if you can trust
timestamps, you can always try this:

https://github.com/lukehutch/fast-classpath-scanner#8-detecting-changes-to-classpath-contents-after-the-scan

This doesn't actually return the last-modified timestamp, so it will only
detect changes between calls to .scan(), but you could add a method that
returned the last-modified timestamp, and store it in a persistent store
somewhere, so that change detection worked across restarts.

[...]

PR #11, opened 2015-07-24
Fix for issus #10 Whitelisting doesn't work

wuminorb:
When scan root dir for classes. The relative path will be “/”.But the
whitelisted path start without “/“.So, fix it with keeping recursing
for “/“.Please check if it’s suitable.


lukehutch:
I believe this is the correct fix. Thank you for the patch! I'll make another release now.

PR #12, opened 2015-07-25
Test cases for package whitelisting and subclasses. Some are broken.

caillette:
I applied Wuminorb's patch locally and it makes some test cases pass but not all of them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -128,6 +128,15 @@
 
 	</build>
 
+	<dependencies>
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
+			<version>4.12</version>
+		</dependency>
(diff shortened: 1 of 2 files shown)


lukehutch:
Hi, can you please test again with the latest version I posted to github, and with the test fixes I described above (use FastClasspathScannerTest.class, not FastClasspathScanner.class).

I also added my own test class, in place of this pull request, that is more extensive, but please feel free to add additional tests if you would like. Thanks!

caillette:
I've reforked everything so ignore this pull request, there is a new one coming.

lukehutch:
Sorry for all the churn. I'm done working on this for today, so you can sync again with HEAD if needed, and it hopefully won't change out from under you again!

PR #13, opened 2015-07-25
FastClasspathScannerTest improvement

caillette:
There are two broken tests: scan and scanFilePattern.


caillette:
I've checked that all of my previous tests do appear in FastClasspathScannerTest.

PR #17, opened 2015-07-30
Support Java 7

danhaywood:
I'd like to use your project (great work by the way) in Apache Isis, and remove our dependency on Reflections and (from some spiking I've done) reduce our start-up times.

Unfortunately in Apache Isis we are still committed to supporting Java 7, whereas you are building the project for Java 8-only.  

I decided to fork the project to see if it could be backported, and seemingly it can with only a couple of changes to test files.  Other changes in this PR are just some configuration to maven-javadoc-plug (to suppress errors thrown up by docLink) and update ..gitignore to ignore IntelliJ project files.

I expect we'll move Isis up to Java 8 within the next 12 months (certainly not beyond the release of Java 9 in Sept 2016) so perhaps you'd consider supporting Java 7 until then?

ALTERNATIVELY, perhaps you'd be happy to just push this branch out to Maven central as a one-off, eg with a version "1.4.1.jdk7" or similar?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ lib/
 *~
 hs_err_pid*
 
+*.iml
 .project
 .metadata
 .classpath
--- a/pom.xml
+++ b/pom.xml
@@ -52,8 +52,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.1</version>
 				<configuration>
-					<source>1.8</source>
-					<target>1.8</target>
+					<source>1.7</source>
+					<target>1.7</target>
 				</configuration>
 			</plugin>
 
@@ -93,6 +93,9 @@
 						<goals>
 							<goal>jar</goal>
 						</goals>
+						<configuration>
+							<additionalparam>-Xdoclint:none</additionalparam>
+						</configuration>
 					</execution>
 				</executions>
 			</plugin>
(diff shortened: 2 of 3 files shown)


lukehutch:
Hi Dan, I'm glad it works for you, I appreciate the patch, and I agree, some of these Reflections-type libraries are massive overkill for many projects, and can incur a big speed penalty!

I intended to not use Java 8 features in FastClasspathScanner itself once I realized it carries an overhead of 30-40ms the first time you use a lambda or similar feature. The test code definitely doesn't need to be based on Java 8, so I'm happy to drop down the minimum version requirement.

I just pushed out version 1.6.0 to Maven with your patches incorporated. Please test and let me know.

Feel free to mention the benefits of FastClasspathScanner on the Apache Isis site if you'd like :-)

danhaywood:
Thanks for turning this round so quickly, Luke.

I've added a note thanking you to our "thanks" page [1]

Cheers
Dan

[1] http://isis.apache.org/more-thanks.html

lukehutch:
You're welcome, Thanks again for the patch! BTW, not a big deal, but my
name is Luke Hutchison, not Luke Hutch :)

danhaywood:
Apols; name fixed!

cheers
Dan

lukehutch:
Hi Dan, just a quick question on your request to drop the Java dependency from Java 8 to Java 7. I have several classes marked with the FunctionalInterface annotation; was this a problem for you when running under Java 7? Did you just remove these annotations in your own local copy, or were you using JDK 8 in Java 7 compatibility mode?

danhaywood:
The latter... I compiled with JDK 1.8 but with the maven-compile-plugin
configured with target (and, IIRC, source) set to 1.7.

Issue #20, opened 2015-08-19
doesn't find classes in jar file manifest's classpath


tsujamin:
I've been using fast-classpath-scanner to locate classes in my classpath supporting a particular interface (allowing users to add functionality to my project). The classes would be found when the application was invoked like so:
`java -cp base-application.jar:base-extension.jar:user-extension.jar MainClass`

I've recently added a Class-Path line to base-application's manifest (adding base-extension.jar), meaning I should be able to execute the application like so:
`java -cp user-extension.jar -jar base-application.jar`

However, fast-classpath-scanner appears to no longer locate classes contained in the Class-Path attribute. The classes don't appear to be loaded at all. I've tested the old and the new version with verbose classloading enabled (both have the same classpath): http://pastebin.com/KS5kRqqr

lukehutch:
Thanks for the report. I think you're saying that FastClasspathScanner doesn't detect classes added using the Class-Path line in a jar's manifest file. Are you using the latest version in git? I checked in some additional classpath detection methods recently, but haven't made an additional release yet.

tsujamin:
Yep that's what I meant. I'll pull from git now and test it

tsujamin:
No luck on HEAD. I had to add " -Dmaven.test.skip=true" when I built fast-classpath-scanner because all the tests were failing on my machine...

tsujamin:
The case where I specify the classpath at commandline (instead of with the Class-Path jar line) still works on HEAD

lukehutch:
Weird. I'm looking at this now, and I don't see entries from either the "-cp" switch or from the Class-Path manifest line in either the java.class.path property, or in the list of URLs returned by the ClassLoader. I have no idea how to read these other classpath entries!

tsujamin:
Apparently they just get loaded the same as the jar that refers to them:
"In general, classes referenced by a JAR-class-path entry are found as though they were part of the JAR file. The JAR files that appear in the JAR-class-path are searched after any earlier class path entries, and before any entries that appear later in the class path." - https://docs.oracle.com/javase/8/docs/technotes/tools/findingclasses.html

So maybe they are prefixed weirdly or something, I'm no expert on HotSpot's classloader implementaiton ;)

tsujamin:
There's a wrapper class for manifests which could be used, but that would probably require attempting to load the manifest for each jar in the classpath. Not sure how expensive that would be.

[...]

PR #21, opened 2015-08-19
Added support for JAR manifest "Class-Path" attribute

tsujamin:
JAR files added to the classpath by inclusion in the "Class-Path"
attribtue of a JAR manifest are not found using current method. They are
now retrieved by accessing a private field within the system class
loader.

Is this a method you want to use to support the Class-Path attribute? If so, I'm guessing I'd probably need to perform this for all classloaders, not just the system class loader.


lukehutch:
Thanks for doing the work on this. However, see my comment on Issue #20. (If the hacky but less-hacky-than-reflection method of reading the manifest files works, I'd prefer that to using hidden Sun internal methods and fields...)  Let me know if the latest HEAD works for you.

PR #22, opened 2015-08-21
Add support for scanning and retrieving by meta annotation

stickycode:
I make lots of use of Meta annotations, and anyone that has written an annotation probably has too ElementType, Inherited, Documented etc. 

Was straight forward to add and easy to test, let me know if there is anything to improve.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,4 +1,5 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 
 	<groupId>io.github.lukehutch</groupId>
@@ -137,6 +138,12 @@
 			<artifactId>junit</artifactId>
 			<version>4.12</version>
 			<scope>test</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>org.assertj</groupId>
+			<artifactId>assertj-core</artifactId>
+			<version>3.1.0</version>
 		</dependency>
 	</dependencies>
 
--- a/src/main/java/io/github/lukehutch/fastclasspathscanner/FastClasspathScanner.java
+++ b/src/main/java/io/github/lukehutch/fastclasspathscanner/FastClasspathScanner.java
@@ -507,6 +507,22 @@ public class FastClasspathScanner {
         return classGraphBuilder.getNamesOfClassesWithAnnotation(annotationName);
     }
 
+    public List<String> getNamesOfClassesWithMetaAnnotation(final Class<?>... annotations) {
+        String[] names = new String[annotations.length];
+        for (int i = 0; i < annotations.length; i++) {
+            Class<?> annotation = annotations[i];
+            if (!annotation.isAnnotation())
+              throw new IllegalArgumentException("Class " + annotation.getName() + " is not an annotation");
+
+            names[i] = annotation.getName();
+        }
+
+        return classGraphBuilder.getNamesOfClassesWithMetaAnnotation(names);
+      }
+
+    public List<String> getNamesOfClassesWithMetaAnnotation(String... metaAnnotation) {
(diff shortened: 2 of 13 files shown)


lukehutch:
Thanks for this, the patch looks pretty good.

So in getNamesOfClassesWithMetaAnnotation(), you are returning both the annotations that have the meta annotation, and the classes that are annotated with annotations that have the meta annotation?

I'm curious about use cases, can you elaborate what you're using it for? I should add some brief docs about this too.

stickycode:
Was supposed to be classes annotated with meta annotation and classes
annotated with meta annotated annotations, which tests should prove as it
says contains exactly.

Use case is building container independent components, in particular
dynamic guice modules based on classpath.

lukehutch:
A quick scan of the code indicates that this patch may be incomplete: I
think you need to build a DAG (well, technically a tree) of
meta-annotations, since you could have a meta-annotation on a
meta-annotation on a meta-annotation, and I think your code will only
handle one layer deep of meta-annotations.

If that's true, search for DAGNode in ClassGraphBuilder and copy the
patterns for interfaces and classes, forming another DAG for annotations.
(There are several places that will need changes/additions.)

Then you'll need a unit test that tests more than one level of
meta-annotation. (Also make sure that the compiler doesn't happily compile
a cycle of meta-annotations, i.e. A meta-annotating B, and B
meta-annotating A, otherwise we need to check for cycles when adding to the
DAG too.)

stickycode:
Only one layer is allowed, although not strictly specified that is the
convention.

lukehutch:
Do you have a reference in a spec (or recommended best practices doc or
something) for that convention? If we're going to add this, I want to make
sure it's complete and correct.

stickycode:
East to test, see if Java supports one of its meta annotations on a meta,
I'm 95% that Its just one level. I'll right a test tonight, just working
the garden right now.

stickycode:
See test I just committed to my fork wi ll pr later

lukehutch:
Pulling in your PR, might add a few more comments and things, but I appreciate the contribution!

lukehutch:
I did some re-working of the patch, and wrote/modified some tests. I want to make sure I'm understanding what meta-annotations are though and how they're supposed to work. Can you please pull HEAD and double-check all the tests are testing the correct semantics?

Specifically, my understanding is that:
(1) The API should return a class given a meta-annotation if either the meta-annotation annotates an annotation that annotates the class, or if the meta-annotation directly annotates the class;
(2) If a meta-annotation M does not directly annotate a class C, but M annotates an annotation A that annotates C, then C sees A as an annotation (using Class.getAnnotation()), but not M.

I tweaked the API a bit so now you can find classes that have either "all of" or "any of" a list of annotations.

lukehutch:
Sorry, I just thought about this some more, and I have more questions :-)

My understanding is that the Java reflections API doesn't have a way of directly reading meta-annotations. If you can only read direct annotations for a given class using the Java reflections API, and not meta-annotations, then it's up to you to read the annotations on the annotations.

If that's true, then it is also true that meta annotations can be transitive, because you can keep reading annotations on annotations on annotations until you get to a "top" annotation that doesn't have its own annotations.

What leads you to believe that annotations only apply two levels deep? What tool or API are you using meta-annotations with that allows meta-annotations but not meta-meta-annotations? (I guess I never understood how your original tests were proving that meta-meta-meta-annotations were not possible, using only the reflections API...)

stickycode:
The retention, element type meta annotations that are part or Java. The
test showed that the retention annotation was only honoured to one level.
If it was further then the annotation would be retained.

lukehutch:
Ah, I see. Makes sense, thanks!

lukehutch:
I found this:

https://github.com/dblevins/metatypes/

The fact that Java doesn't carry its own annotation properties more than
one level is just due to there being no need. You can construct a deep
graph with annotations, and there's no reason you can't use more levels of
the graph than Java does itself.

I think I am going to do a full DAG implementation of this, so you can have
meta^N-annotations. It should be a strict superset of Java's functionality,
so it shouldn't break anything you're doing. Will let you know when I'm
done so you can test.

stickycode:
All good, will work for me either way. Happy to test. Let me know.

lukehutch:
Hi Michael -- I just pushed a major update to the meta-annotations support. The API changed (method names now just say "Annotations", not "MetaAnnotations", and the "Annotations" methods work with both annotations and meta-annotations). The transitive closure graph of meta-annotations is determined, and cycles in the annotation graph are handled correctly. I would appreciate your testing! (Sorry to change the API on you.)

Thanks again, it was an interesting problem to work on.

stickycode:
Awesome I'll have a look in a few hours.

stickycode:
Works for me, don't have any micro bench marks so can't comment on
performance.

Used head and 1.9.0

lukehutch:
Great, thanks for testing. Happy to know I managed to create a generic
superset of what you needed!

PR #54, opened 2016-06-16
Fixing squid:S1132 - Strings literals should be placed on the left side when checking for equality

apxeolog-df:
This pull request is focused on resolving occurrences of Sonar rule 
squid:S1132 - “Strings literals should be placed on the left side when checking for equality”. 
You can find more information about the issue here: 
https://dev.eclipse.org/sonar/rules/show/squid:S1132
Please let me know if you have any questions.
Artyom Melnikov.


lukehutch:
I'm always careful to check that strings cannot be null when writing equality tests, and getting an NPE therefore serves as a sanity check (if string literals are on the LHS, you can get silent failures that propagate further), but I guess it is good practice to write things this way. Thanks for the PR!

PR #55, opened 2016-06-17
Fixing squid:S1118 and squid:S1871

apxeolog-df:
This pull request is focused on resolving occurrences of Sonar rules:
squid:S1118 - "Utility classes should not have public constructors".
You can find more information about the issue here: 
https://sonarqube.com/coding_rules#q=squid:S1118
squid:S1871 - "Two branches in the same conditional structure should not have exactly the same implementation".
You can find more information about the issue here: 
https://sonarqube.com/coding_rules#q=squid:S1871
Please let me know if you have any questions.
Artyom Melnikov


lukehutch:
Thanks, I added the private constructors, but the other changes are not very future-proof -- I may need to parse out more of the classfile in future, and combining unrelated fields, or fields of different type, that happen to be encoded as the same number of bytes in the classfile, doesn't seem like the ideal thing to do.

PR #68, opened 2016-07-23
Rename AutoCloseableExecutorService, shutdown executor immediately, e…

johnou:
…xtract thread factory into util class.

### DIFF
--- a/src/main/java/io/github/lukehutch/fastclasspathscanner/FastClasspathScanner.java
+++ b/src/main/java/io/github/lukehutch/fastclasspathscanner/FastClasspathScanner.java
@@ -35,10 +35,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import io.github.lukehutch.fastclasspathscanner.classloaderhandler.ClassLoaderHandler;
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/lukehutch/fastclasspathscanner/FastClasspathScanner.java
```diff
@@ -178,7 +177,6 @@ private synchronized ScanSpec getScanSpec() {
                 try {
                     // for (int i = 0; i < 10000000; i++) { // TODO:
```
johnou: An interesting todo :)

src/main/java/io/github/lukehutch/fastclasspathscanner/FastClasspathScanner.java
```diff
@@ -219,32 +217,22 @@ public synchronized static final String getVersion() {
...
-                    });
+                    threadFactory);
         }
 
         /** Shut down thread pool on close(). */
         @Override
         public void close() {
             try {
-                shutdown();
+                shutdownNow();
```
johnou: from what I could see all operations inside the try catch for AutoCloseableExecutorService are blocking so this should be safe?

johnou:
@lukehutch finally managed to look at the master branch using a monitor, awesome work!

src/main/java/io/github/lukehutch/fastclasspathscanner/FastClasspathScanner.java
lukehutch: Oh yeah, this was for profiling / performance testing :)

src/main/java/io/github/lukehutch/fastclasspathscanner/FastClasspathScanner.java
lukehutch: That was the intent, yes.

lukehutch:
Thanks for your work on this!

PR #82, opened 2016-09-29
Adding OSGi metadata generation to distribution

bertramn:
Hi Luke,

I would like use your classpath scanner in a OSGi compliant runtime like Apache Felix but that would require some additional metadata to be added to the assembled jar file. Would you be up for merging this PR which adds the required OSGi descriptor to the MANIFEST.MF file? It should not have any impact on any current use of the jar and I have tested the maven build.

Thanks,
Niels


lukehutch:
Merged, thanks so much for your contribution! I'll push out a new release soon.

lukehutch:
Released in version 2.0.4.

PR #84, opened 2016-10-03
Adding OSGi metadata to existing projects without changing the packag…

johnou:
…ing type.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,6 @@
 	<artifactId>fast-classpath-scanner</artifactId>
 	<version>2.0.4-SNAPSHOT</version>
 	<name>Fast Classpath Scanner</name>
-	<packaging>bundle</packaging>
 
 	<description>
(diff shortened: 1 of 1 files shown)


johnou:
@bertramn PTAL.

bertramn:
Looks good, dropped it into Karaf 4 and budnle comes up nicely. The only optimisation would be to consider if any of the packages can be considered "internal" use and excluded from export. But by the looks of it there is at least 1 class one would typically use in each package (not that this matters much TBH).

johnou:
@lukehutch I think we are good to go :)

lukehutch:
Thanks, @johnou . @bertramn , you could exclude the utils subpackage if you would like, those classes are all for internal use. I suspect you don't feel too strongly about this based on your comment "not that this matters much TBH", so I'll merge this PR now -- feel free to submit another one if you want to exclude utils from export.

lukehutch:
Released in version 2.0.4.

PR #85, opened 2016-10-05
Fixes classpaths for Wildfly managed deployments.

msteur:
This fixes the classpath when deployed through the web interface instead of the hot deployment folder.
The previous method would return a virtual path that could not be resolved.
This is probably where this tread hinted at: https://github.com/lukehutch/fast-classpath-scanner/issues/37

This has only been tested with war files on Wildfly 10.0.0-Final on Linux.
It works when deployed by the web interface and when copied to the deployment folder. (By Eclipse)

### DIFF
--- a/src/main/java/io/github/lukehutch/fastclasspathscanner/classloaderhandler/JBossClassLoaderHandler.java
+++ b/src/main/java/io/github/lukehutch/fastclasspathscanner/classloaderhandler/JBossClassLoaderHandler.java
@@ -28,6 +28,7 @@
  */
 package io.github.lukehutch.fastclasspathscanner.classloaderhandler;
 
+import java.io.File;import java.io.FileFilter;
 import java.lang.reflect.Array;
 
 import io.github.lukehutch.fastclasspathscanner.scanner.ClasspathFinder;
@@ -40,6 +41,13 @@ import io.github.lukehutch.fastclasspathscanner.utils.ReflectionUtils;
  * https://github.com/jboss-modules/jboss-modules/blob/master/src/main/java/org/jboss/modules/ModuleClassLoader.java
  */
 public class JBossClassLoaderHandler implements ClassLoaderHandler {
+    private static final FileFilter FILE_FILTER = new FileFilter() {
+        @Override
+        public boolean accept(File pathname) {
+            return pathname.isFile() && pathname.exists();
+      }
+    };
+  
     @Override
     public boolean handle(final ClassLoader classloader, final ClasspathFinder classpathFinder, final LogNode log)
             throws Exception {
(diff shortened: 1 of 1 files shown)


msteur:
I just checked out the reflections project, and they seem to use the same approach as I've used.

lukehutch:
Thanks for this PR. I have a question though: you're adding all the sibling files of each classpath element inside your "if (file != null)" block -- is this the right thing to do in the general case? Is there a more reliable way to find the library jars inside a Wildfly jar, e.g. by looking in the manifest or something? (Adding all siblings of all Wildfly classpath entries seems a bit dangerous, you could include all sorts of weird stuff into the classpath in non-standard situations...)

lukehutch:
And actually, the Vfs logic you linked to is quite different from yours (and more along the lines of what I would expect to work, without adding other random siblings of classpath entries into the classpath):

```java
                Object content = url.openConnection().getContent();
                Class<?> virtualFile = ClasspathHelper.contextClassLoader().loadClass("org.jboss.vfs.VirtualFile");
                java.io.File physicalFile = (java.io.File) virtualFile.getMethod("getPhysicalFile").invoke(content);
                String name = (String) virtualFile.getMethod("getName").invoke(content);
                java.io.File file = new java.io.File(physicalFile.getParentFile(), name);
                if (!file.exists() || !file.canRead()) file = physicalFile;
                return file.isDirectory() ? new SystemDir(file) : new ZipDir(new JarFile(file));
```

lukehutch:
@msteur I just pushed a change to the code so that it operates the same as Reflections; can you please test this and let me know if it solves the problem? If not, please let me know the answers to the questions in the comments above. Thanks!

msteur:
I've tested it and it works.
It seems to be a better solution than mine indeed.

I think the magic is in this part, I would have expected this to be the same file, but it's not.

```Java
final File physicalFile = (File) ReflectionUtils.invokeMethod(root, "getPhysicalFile");
String name = (String) ReflectionUtils.invokeMethod(root, "getName");
File file = new java.io.File(physicalFile.getParentFile(), name);
```

The files I iterated through would only result in adding all libraries from my war file, so no weird stuff.

lukehutch:
Thanks for testing. When you say "I've tested it and it works", do you mean the version in git master, or your PR?

Can you please look at the `.verbose()` output from the version in git master, and see if the nested libraries are properly found? They should already be in the classpath, added by some Wildfly ClassLoader, and they shouldn't need to be explicitly included by including sibling files as you have in your PR.

If the nested libraries are not being found, then we may need to read some other metadata file from the jarfile in the Wildfly case to find the location of the library jars. (We should solve this in a generic way, in case the packaging structure changes.)

msteur:
I meant that I tested the master, I had already tested my pull request. ;)
I can confirm that it picks up the nested libraries, because otherwise my tests would have failed. Part of the files I'm scanning for are in separate libraries.

The nested loop that I used would pull those in, but that's not required with the current setup.

lukehutch:
@msteur great, thanks for testing, and I'll close this pull request then, but thanks so much for sending it in. It's always nice when somebody sends in not just a bug report of feature request, but actual code :-)  Your pointer to Reflections was really useful too, I will probably use that to build support for a couple more URL types.

I might wait a couple of weeks before pushing out a new version, to reduce version number churn.

PR #98, opened 2017-01-10
Whitelisting JARs should not prevent scan directories.

michaldo:
Applications consist of 2 parts: third party code and my code. I can
speed up classpath scanning by providing list of JARs I want to scan.
If I want scan my own artifacts, I have a problem launching application
from Eclipse. If my own artifact is opened as a project, Eclipse will
attach to classpath folder "target" of the artifact, not JAR. 

Finding which success in Spring Boot standalone application or plain WAR
application on Tomcat will fail in application running from Eclipse.

I don't want modify scanner whitelist for running from Eclipse. A
solution is revert current default and scan classpath directories even
with whitelisted JARs.

I expect performance impact is low because large applications has most
of code packed into JARs and directories are little comparing to JARs.
However, previous default is still possible to configure by
specification "-dir:".

Specification "jar:" (JARs only) is still respected and disables
directory scanning.

### DIFF
--- a/src/main/java/io/github/lukehutch/fastclasspathscanner/scanner/ScanSpec.java
+++ b/src/main/java/io/github/lukehutch/fastclasspathscanner/scanner/ScanSpec.java
@@ -271,6 +271,21 @@ public class ScanSpec {
                             }
                         }
                     }
+                } else if (spec.startsWith("dir:")) {
+                	spec = spec.substring(4);
+                	if (!spec.isEmpty()) {
+                		if (log != null) {
+                            log.log("Only \"dir:\" is supported, \"dir:" + spec + "\" is ignored");
+                        }
+                	} else if (blacklisted) {
+                        // Specifying "-dir:" blacklists all directories for scanning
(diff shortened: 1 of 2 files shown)


lukehutch:
Oops, I didn't read this closely enough, I thought this was a bug report, not a PR! I actually implemented almost exactly the same change, with the same syntax (using `-dir:` and `dir:`). Great minds think alike :-)  The only major difference is that `jar:` does not exclude directories automatically in my version, you have to manually exclude dirs if you don't want them scanned.

The docs are updated here:

https://github.com/lukehutch/fast-classpath-scanner/wiki/2.-Constructor#scan-spec

Please let me know if this works for you. I'll close this PR, since it does more or less the same thing as what I committed (although I also used this as an opportunity for some code cleanup and simplification). Sorry for the duplication of work, and thanks for submitting a PR!

michaldo:
Hi Luke, 
my PR is the same idea as your commit, so closing is OK.

PR #113, opened 2017-05-10
add bundle itself as a classpath element if it does not have Bundle-Classpath

qalex:
If an OSGi bundle does not have a Bundle-Classpath statement then it'll be simply skipped by the FelixClassLoaderHandler. This simple fix adds the bundle itself as a classpath element for scanning.

Can't say if this is a universal solution, but works for me. May be we should always add the bundle, even if it does have a Bundle-Classpath?


lukehutch:
@qalex Thank you for tracking down the cause, and thanks so much for submitting this pull request!

I don't use OSGi, but in general handling non-native classloaders is a long tail of special cases like this, and this patch looks good to me.

PR #121, opened 2017-06-03
refactor maven project setup for proper test execution

jjYBdx4IL:
and move release plugins to -Prelease profile such that "mvn clean install" should work everywhere.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,19 +52,6 @@
 	<build>
 		<plugins>
 			<plugin>
-				<!-- This plugin doesn't work inside pluginManagement, for some reason -->
-				<groupId>org.sonatype.plugins</groupId>
-				<artifactId>nexus-staging-maven-plugin</artifactId>
-				<version>1.6.3</version>
-				<extensions>true</extensions>
-				<configuration>
-					<serverId>ossrh</serverId>
-					<nexusUrl>https://oss.sonatype.org/</nexusUrl>
-					<autoReleaseAfterClose>true</autoReleaseAfterClose>
-				</configuration>
-			</plugin>
-
-			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-source-plugin</artifactId>
 				<version>2.2.1</version>
@@ -93,33 +80,6 @@
 						</configuration>
 					</execution>
 				</executions>
-			</plugin>
-
-			<plugin>
-				<groupId>org.apache.maven.plugins</groupId>
-				<artifactId>maven-gpg-plugin</artifactId>
-				<version>1.5</version>
-				<executions>
-					<execution>
-						<id>sign-artifacts</id>
-						<phase>verify</phase>
-						<goals>
-							<goal>sign</goal>
-						</goals>
-					</execution>
-				</executions>
-			</plugin>
-
-			<plugin>
-				<groupId>org.apache.maven.plugins</groupId>
-				<artifactId>maven-release-plugin</artifactId>
-				<version>2.5</version>
-				<configuration>
-					<autoVersionSubmodules>true</autoVersionSubmodules>
-					<useReleaseProfile>false</useReleaseProfile>
-					<releaseProfiles>release</releaseProfiles>
-					<goals>deploy</goals>
-				</configuration>
 			</plugin>
 
 			<plugin>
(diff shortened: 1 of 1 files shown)


lukehutch:
Hi @jjYBdx4IL, thanks for the PR. I really don't know much about Maven, so could you please explain what this patch does (or why it wasn't working for you in the previous form)?

Also, several of the files appear to have had their line ending changed from LF to CRLF or something, e.g. IssuesTest.java. Can you please update the PR to use LF instead?

jjYBdx4IL:
the separation of the actual module and the tests is necessary because with your current setup, "mvn install" is broken and does not run on my machine.

The second thing that is breaking "mvn install" is the fact that your release-specific stuff is interfering with normal development and needs to be placed into a release profile. Everyone uses "mvn install", but nearly noone need to deploy stuff to maven central. Therefore simply specify "-Prelease" when publishing to central to make it work by default for everyone else.

The CRLF stuff is sadly because IDEs expect files to be in the OS' native "linefeed encoding" and git always messes around with it on checkout and on checkin, though I don't understand yet why CRLF ended up in the repository because git should transformed everything into LF. But that was probably because eclipse git didnt recognize the corresponding files as text files and eclipse is too stupid to handle mline endings according to the already present line endings in the file and ends up writing files with mixed line endings. Just one of many issues that never got fixed by simply establishing an UTF-8/LF standard for everything.

jjYBdx4IL:
Can you point me to the files with changed line-feeds? I can't find them. Maybe just add them to .gitattributes and append text=auto. See https://help.github.com/articles/dealing-with-line-endings/

jjYBdx4IL:
I checked IssuesTest.java on a linux box. No CRLF there. Working as expected.

jjYBdx4IL:
You can also remove the parent declaration in dist/pom.xml to avoid publishing parent/tests modules to central.

lukehutch:
@jjYBdx4IL maybe it wasn't CALF issues then, it could have been a tabs/spaces issue or some other formatting change. I simply eyeballed the diff, and there were several files where it looked like the whole file changed, but I couldn't see any visible change. Take a look at GitHub's diff view, if you could please! 

Thanks for explaining the changes. I'm traveling right now, but will pull in this PR hopefully in the next couple of days.

johnou:
Seems a little strange to split tests into another module, I don't see how your main concern couldn't be addressed with simple changes to the root pom.xml.. Furthermore "fixing" the line endings should be in a separate pull request, it adds far too much noise and reviewing the actual changes a headache. Tbh I would dislike reformatting the entire code base for no good reason, it would destroy the git "blame" in IDEA.

lukehutch:
@johnou thanks for the feedback, I agree, however I know so little about Maven that I don't know how to fix the reported issue. @jjYBdx4IL, can you see a simpler way to accomplish this?

jjYBdx4IL:
this should be better. it assume Java 8 for normal operations and does your backwards compat stuff only when releasing.

But I suggest releasing two modules: the core Java 7 compat module, and an extension module for Java 8 features, ie. fast-classpath-scanner-java8 or something. Apart from that, maybe you can find something that simply checks Java 7 compatibility instead of using the deprecated Java 7 SDK/JRE. It basically boils down to checking whether your code uses only Java 7 APIs.

lukehutch:
@jjYBdx4IL Thanks for the update. Just so I'm clear on what you're trying to accomplish here, you want to compile the library with JDK 8 in Java 8 mode when using "mvn install"? What's the issue with using the JDK 8 compiler in Java 7 compatibility mode? (I definitely need to compile using only Java 7 bytecodes, since I have a lot of users stuck on Java 7.)

jjYBdx4IL:
the profile is needed because of the pgp plugin. if a user has no pgp configured, it fails. The test fails because of no specific reason. Of course you can use jdk7 mode by default, but some of the tests will fail then. Also, you should use an outdated jdk7 to perform your release compilation because if you don't, you don't actually verify that you aren't using some newer jdk8 APIs.

jjYBdx4IL:
actually I was trying to fix a bug, but I could not reproduce it. at some point in your code you are checking classpath URLs for safe characters but not securing against ArrayIndexOutOfBoundsException.

jjYBdx4IL:
I think it was here: 

https://github.com/lukehutch/fast-classpath-scanner/blob/44941997b723d05c5d29f165d4e4d0b4f3ba810c/src/main/java/io/github/lukehutch/fastclasspathscanner/utils/URLPathEncoder.java#L73

lukehutch:
@jjYBdx4IL Yes, that's clearly something that could trigger ArrayIndexOutOfBoundsException -- thanks for the report, fixed.

lukehutch:
@jjYBdx4IL these changes to the pom file broke pushing to Maven Central (it fails silently, I believe) -- so the last couple of releases aren't published. Any idea why?

johnou:
@lukehutch try the bottom snippet http://maven.apache.org/maven-release/maven-release-plugin/examples/perform-release.html

lukehutch:
@johnou thanks for the suggestion!

PR #125, opened 2017-06-16
auto-activate release profile when doing release

jjYBdx4IL:
That's what I do in my own projects.


jjYBdx4IL:
ie.

     		<profile>
      			<id>release</id>
     +			<activation>
     +				<property>
     +					<name>performRelease</name>
     +					<value>true</value>
     +				</property>
     +			</activation>

jjYBdx4IL:
but yeah, you should probably also use the hint you pointed out, ie. tell the release plugin which profile to use. Can't say for sure because I can't test that stuff here.

lukehutch:
Thanks for this..

lukehutch:
@jjYBdx4IL It looks like this fixed it -- released as 2.2.1, and it made it to Sonatype this time.

jjYBdx4IL:
Cool.

PR #126, opened 2017-06-28
Fixed WASLiberty classloading for version 2.4.0.

RaimondKempees:
I suspect that since removing fallback classloading is removed, the appClassLoader isn't loaded anymore. If that is the case, I now get it from the ThreadContextClassLoader in WAS Liberty.


lukehutch:
Published in version 2.4.1.

RaimondKempees:
Awesome! Thanks!

Just tested with 2.4.1 and it's working like a charm.

If I encounter more exotic application containers with custom classloaders,
I'll make sure to keep contributing. :)

lukehutch:
That would be awesome. Before I started this project, I had no idea to what
degree classloading was the wild west! I definitely could not make the
coverage of cases complete without contributions like yours.

PR #132, opened 2017-07-31
ClassLoaderHandler implementation for Traditional (not Liberty) Websphere

sbespalov:


lukehutch:
@sbespalov I added the MIT license header to your source file (the entire project is under the MIT license) -- please confirm that this is OK with you.

sbespalov:
yes, of course
Thanks

RaimondKempees:
@lukehutch, @sbespalov I can confirm this works on Websphere Traditional 8.5.5 and Websphere Traditional 9.0.0.4 when I am using 2.4.6-SNAPSHOT. Given https://github.com/lukehutch/fast-classpath-scanner/issues/130, I am unsure whether we should add com.ibm.ws.bootstrap.ExtClassLoader and com.ibm.ws.classloader.ProtectionClassLoader

I did create a PR for this ()
Finally, a good primer is here: http://www-01.ibm.com/support/docview.wss?uid=swg27023549&aid=1 :)

lukehutch:
@RaimondKempees Thanks for this, and for the informative link, I'll link to that from the docs. As far as Ext and Protection classloaders, I think we should probably add them in, since I'm trying to bring in all URLs from all classloaders as the default policy. Better to scan more than needed rather than miss things that are needed.

RaimondKempees:
Fair enough :)

PR #133, opened 2017-08-18
Support filtering scan results by classloader

notsyncing:
This pull request adds a parameter to almost all methods in `ScanResult`, allowing for getting classes only from one classloader. Our use case is:

To support dynamic plugin loading and unloading, I use two classloaders: a `parentClassLoader` and a `childClassLoader`. the `parentClassLoader` contains class `A` and `B`, while the `childClassLoader` loads plugin class `C` in another jar. This jar is not in the classpath of `parentClassLoader`, only in the `childClassloader`. Class `C` extends class `B`, and class `B` extends class `A`. `A` and `B` are not in the same jar, but they are both loaded by `parentClassLoader`.

Then, I need to scan only the loaded jar for all subclasses of `A` and run some initial methods on them. I had tried following methods:

```java
ScanResult result = new FastClasspathScanner()
    .overrideClasspath("<URL of jar containing class C>").scan(); // Method 1
//new FastClasspathScanner().addClassLoader(childClassLoader).scan(); // Method 2
//new FastClasspathScanner().overrideClassLoaders(childClassLoader).scan(); // Method 3

result.getNamesOfSubclassesOf(A.class); // Get all subclasses of A
```

When using method 1, it cannot find class `C`, but if I call `getNamesOfSubclassesOf(B.class)`, it can find class `C`. I plot the class graph with `generateClassGraphDotFile`, and found that class `B` is not connected to class `A` in this case.

Then I used method 2 and 3, they can find class `C`, but also contain classes from other jars in the classpath, while I just want results from the jar containing class `C`.

So finally I decided to modify this library, and get results only from the specified classloader. It works, but I'm not sure if it was the appropriate way to do so, because I thought I've modified too many things. If you need a test project to reproduce the problem in method 1, please let me know.

Finally, thanks for creating such an amazing library, it runs very fast :)


lukehutch:
Hi @notsyncing, sorry for the delay, I have been away for several days. Thanks for this PR, and I'm glad the library works for you with your modifications!

I recently had to modify FCS to keep track of the classloader that loaded each class that matched scan criteria, so that `Class.forName()` could be called with the specific classloader parameter for any given named class. Consequently, I'm surprised that what you're looking for didn't work out of the box, particularly using your Method 2, which would have been my recommendation to achieve what you describe. I'd almost rather get Method 2 working properly, since this is supposed to work without the extra plumbing you added. The only case I can think of that shouldn't work is if you want to have two different instances of a given class loaded in different classloaders (which therefore couldn't share a parent, so that there is no caching).

I'll have to look in more detail at the changes you have made, and think about why the current head version doesn't work for your Method 2. I'm guessing I should be able to take the unit test in your PR and modify it to try to scan with the child classloader using your Method 2, in order to reproduce the problem?

notsyncing:
Thanks for the reply. I think the result in method 2 is correct, because it seems to add `childClassLoader` to the scanning list, and FCS can find classes loaded by `childClassLoader`. What I thought having incorrect results were method 1 and 3. I only want classes from `childClassLoader`, and method 1 gives broken class dependencies, while method 3 still contains classes from other classloaders.

Meanwhile, I'm not having two different instances of a given class. What I want is following:

expected: call `getNamesOfClasses(A.class)`, get only class `C` in result
method 1: got nothing, but can get `C` when call `getNamesOfClasses(B.class)`
method 2: got both class `B` and `C`
method 3: got both class `B` and `C`

The unit test in my PR is just to test the classloader filtering. I'll try to make another unit test to reproduce this problem.

lukehutch:
OK, I'm looking at your `IndirectInheritedClassScanResultTest.java` test.

Method 1 is working as expected -- see the [docs](https://github.com/lukehutch/fast-classpath-scanner/wiki/3.1.-Finding-subclasses-or-superclasses):

    *Important note:* the ability to detect that a class extends another depends upon the entire ancestral path between the two classes being within one of the whitelisted package prefixes. (However, [see here](https://github.com/lukehutch/fast-classpath-scanner/wiki/2.-Constructor#detecting-annotations-superclasses-and-implemented-interfaces-outside-of-whitelisted-packages) for info on "external" class references.)

If you change the code (as currently checked into your branch) to search for `classloaderfiltered.B` rather than `classloaderfiltered.A`, the test succeeds (using Method 1). This is all working as expected.

If you disable Method 1 and enable Method 2, both B and C are found, as you described. If you instead do `.addClassLoader(parentLoader)`, you only get B. Again, this is all working as expected.

Method 3 results are exactly the same as Method 2 results (also as expected, since none of these 3 jars are even in the regular classpath).

So, as far as I can tell, Method 2 already does what you need, without any of your modifications. Is that an incorrect assumption?

Method 1 should probably be discouraged. See [here](https://github.com/lukehutch/fast-classpath-scanner/wiki/1.-Usage#important-note-on-manually-overriding-the-classpath). Method 2 is also not ideal (but is needed for your situation) -- see "Important note" [here](https://github.com/lukehutch/fast-classpath-scanner/wiki/4.-Working-with-nonstandard-ClassLoaders#manually-overriding-classloaders).

notsyncing:
I want only C in the result, but method 2 and 3 both contain class B, which I don't need.

lukehutch:
@notsyncing I see -- the reason you are getting `B` too is that you are delegating to `parentLoader`:

`URLClassLoader childLoader = new URLClassLoader(new URL[] {testJarCUrl}, parentLoader);`

So what is happening is that both `parentLoader` and `childLoader` are being scanned, in that order. If you *only* want class `C` and not `B`, you should disable delegation:

`URLClassLoader childLoader = new URLClassLoader(new URL[] {testJarCUrl}, null);`

However, then you have to search for all individual classes that extend `A` that are outside of the URL(s) handled by childLoader, and individually search for those as possible superclasses of `C`, because of the rule I quoted previously:

> the ability to detect that a class extends another depends upon the entire ancestral path between the two classes being within one of the whitelisted package prefixes.

The reason for this is that FCS cannot search for subclasses of `A` if `A` is not the *direct superclass* of a class in the scanning path. If `B` extends `A` and classes in the scanning path only extend `B`, not `A`, FCS has no way to know that `B` extends `A`, because the external class hierarchy is not scanned.

So, you have two options:

(1) Scan everything, and eliminate `B` from the scan results (this way you can still ask the question, "what extends `A`?")

(2) Run a pre-scan of the "external classses", and ask "what extends `A`?" (getting the answer `{B}`), then scan only `testJarCUrl` (using Method 3 with `null` delegation), and in a for-loop, iterate through all the classes that extend `A` (here just `B`), and find all the classes that extend each class iterated through in the for-loop. Add all matching classes to a `Set`.

```java
    @Test
    public void testLoadFromDifferentClassLoaders() {
        URL testJarAUrl = getClass().getResource("/indirect-dep-test-a.jar");
        URL testJarBUrl = getClass().getResource("/indirect-dep-test-b.jar");
        URL testJarCUrl = getClass().getResource("/indirect-dep-test-c.jar");

        URLClassLoader parentLoader = new URLClassLoader(new URL[] { testJarAUrl, testJarBUrl },
                getClass().getClassLoader());

        URLClassLoader childLoader = new URLClassLoader(new URL[] { testJarCUrl }, null); // null delegation

        ScanResult parentScanResult = new FastClasspathScanner().
                overrideClassLoaders(parentLoader) // Method 3 (parentLoader)
                .scan();

        List<String> extendsAClassNames = parentScanResult
                .getNamesOfSubclassesOf("io.github.lukehutch.fastclasspathscanner.test.classloaderfiltered.A");

        ScanResult scanResult = new FastClasspathScanner()
                .overrideClassLoaders(childLoader) // Method 3 (childLoader)
                .scan();

        Set<String> extendsAInChildLoader = new HashSet<>();
        for (String extendsAClassName : extendsAClassNames) {
            extendsAInChildLoader.addAll(scanResult.getNamesOfSubclassesOf(extendsAClassName));
        }

        assertEquals(1, extendsAInChildLoader.size());
        assertTrue(extendsAInChildLoader
                .contains("io.github.lukehutch.fastclasspathscanner.test.classloaderfiltered.C"));
    }
```


Does that make sense?

notsyncing:
Thanks for your detailed explain. I think I should go for option 1, and this is what my PR actually does (`getNamesOfSubclassesOf` can only get class names, no way to know if it was loaded by the `childClassLoader`). Also, in our use case, the `childClassLoader` must delegate to `parentClassLoader`, or the classes `C` from `childClassLoader` won't be loaded (superclass `B` not loadable from `childClassLoader`).

Option 2 seems like a too heavy work, we may end with scanning it for several hundred times. If we added something like a class `D` which extends `C`, then we would need to recursively scan it, leading to thousands of scans.

Now I decided to do this: first scan without `childClassLoader`, get result `1`, then scan with `childClassLoader`, get result `2`, and finally compare the results and get classes which exists in `2` but not exists in `1`. So we do not need to modify FCS, and I'll close this PR.

Thanks again for your patience of these days!

lukehutch:
@notsyncing Your last suggestion (computing the set difference between the parent and child classloader) seems to be the right one. That requires only two scans (or rather, one scan per parent, and one scan for each child). Just use `parentSet.removeAll(childSet)` to accomplish this. (But, of course, this modifies parentSet, so make a copy of the parentSet each time you want to compute the set difference.)

lukehutch:
@notsyncing one other possibility -- maybe I should add an option to disable classloader delegation...

lukehutch:
@notsyncing  I added `FastClasspathScanner#ignoreParentClassLoaders()` to the master version. Feel free to test this if you would like.

lukehutch:
Note however that disabling parent classloader scanning can cause a mismatch between the scan results and the class loading order, so if a class is defined in both a parent ClassLoader and a child ClassLoader, the result returned by a scan can differ from what would actually be loaded by calling `Class.forName()` or similar.

notsyncing:
Works like a charm with `overrideClassLoaders()`, thank you very much!

In our use case, a class won't be defined in both a parent classloader and a child classloader, so it's not a problem.

lukehutch:
Great, I'm glad you got it working! Sorry that it wasn't straightforward!

PR #139, opened 2017-09-20
Update AbstractSuperClass.java

febielgiva:
Can you please keep the field as private. Thank you


lukehutch:
This causes the test to fail because the assertion should test against `"[private String field]"` -- but not because the scanner can't find the field.

PR #142, opened 2017-09-23
~ fix for issue 141

wuetherich:
I added support for the new module-related constant pool tags to prevent the RuntimeException.

In case that a module-info.class is parsed, I simply return null in readClassInfoFromClassfileHeader (should be ok, as we won't search for module-info classes anyway!?)


lukehutch:
Thanks for the PR and test!

PR #150, opened 2017-10-27
JBossClassLoader: upgrade to jboss modules 1.6

jknack:
- In 1.6.x the existing FileResourceLoader extends from PathResourceLoader which now uses a java.nio.file.Path as `root` object, previous version this was a `java.io.File`
- It is a simply/fallback fix


jknack:
Awesome, when do you think it can be released?

Thanks

lukehutch:
PS please let me know how urgently you would like a release pushed out to Maven Central. If it's not urgent, I will watch for a couple more fixes to reduce version churn.

lukehutch:
@jknack just got your previous comment. Based on that, I can push out a release now.

lukehutch:
@jknack Released in 2.8.1. Thanks again!

jknack:
You rock! Thank you!

PR #173, opened 2018-01-24
Test to recreate issue 172

lukehutch:


lukehutch:
@anthonykr3 many thanks for this test case!

PR #178, opened 2018-02-01
Updated test with class file for for Parameter arity mismatch

anthonykr3:


lukehutch:
@anthonykr3 This should be fixed in 2.12.2, uploading now. Please let me know if you run into other issues. Thanks again for working with me on this!

PR #179, opened 2018-02-01
Test for class giving Result types could not be reconciled error

anthonykr3:
Hi - it's me again. Another strange parameter combination it seems. I haven't looked at it properly as I'm just leaving, but I've managed to recreate the issue the same way.

The test errors with: java.lang.IllegalArgumentException: Result types could not be reconciled: java.lang.Object vs. W


lukehutch:
@anthonykr3 No problem, thanks and keep them coming! Interesting, this one is a case of a generic type needing to be reconciled with java.lang.Object after type erasure. I hadn't thought of that scenario. This should be relatively easy to fix.

PR #180, opened 2018-02-02
A couple more failing cases

anthonykr3:
I've still got a few more failing cases on the latest version I'm afraid.


lukehutch:
@anthonykr3 We're getting closer :-)  I will take a look.

lukehutch:
Fixed in 2.12.4.

PR #181, opened 2018-02-02
A couple more test cases

anthonykr3:
A couple more test cases after testing the most recent fixes


lukehutch:
:) We'll just keep hammering away on this until everything passes!

lukehutch:
OK, I fixed one of them. For the other one, this is an issue with the parameter metadata array in the classfile not having an entry corresponding to any synthetic parameters. I guess that means that synthetic parameters can't have compiler-added annotations. Of course that is not documented anywhere. I can't therefore tell if that's also true of mandated parameters, so I'm going to have to take a guess here, and assume that mandated parameters can in fact be annotated (meaning that the parameter annotations array is the length of the parameter modifier array, minus the number of entries that have the synthetic modifier bit set).

lukehutch:
@anthonykr3 I committed fixes for both of these -- please try them out before I make another release. Thanks!

PR #182, opened 2018-02-05
Added some additional tests

anthonykr3:
I did some more testing this evening and found a few more failing test cases - sorry!


lukehutch:
@anthonykr3 No worries. Nice to know that Kotlin makes such extensive use of the Java type system!

lukehutch:
@anthonykr3 @Alexander-- this confirms that Kotlin is outputting incorrect attributes in the classfile. No wonder I couldn't find the right way to line up parameters in the different attribute types.

In class `net.corda.core.node.services.vault.GenericQueryCriteria$ChainableQueryCriteria$AndVisitor$DefaultImpls` there is a method `visit` with two parameters, whose concrete type is given in the method's type descriptor as `(net.corda.core.node.services.vault.GenericQueryCriteria$ChainableQueryCriteria$AndVisitor, net.corda.core.node.services.vault.BaseQueryCriteriaParser)`. However, the `MethodParameters` attribute (introduced in JDK8) contains only *one* parameter, named `parser`, with corresponding modifier value 0.

The [spec](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.7.24) says:

> The i'th entry in the parameters array corresponds to the i'th parameter descriptor in the enclosing method's descriptor. . . Effectively, this means the parameters array stores information for all the parameters of the method.

Therefore, there should be *two* entries in `MethodParameters`, not one. This is definitely a bug in Kotlin's classfile generation code.

Note however that this applies only to the method descriptor, not to the method signature, as contained in the `Signature` attribute:

> One could imagine other schemes, where entries in the parameters array specify their corresponding parameter descriptors, but it would unduly complicate the MethodParameters attribute.
> 
> The i'th entry in the parameters array may or may not correspond to the i'th type in the enclosing method's Signature attribute (if present), or to the i'th annotation in the enclosing method's parameter annotations. 

I assume that parameters in the `Signature` attribute and the parameter annotations can differ in arity, according to the bits in the parameter modifiers in the `MethodParameters` attribute. However, if the arity of these attributes doesn't match the arity of parameters in the method descriptor, I will have to ignore the parameter modifiers. If the arity of parameters in the method descriptor also doesn't match the arity of parameters in the method signature, there's no way to figure out which method information to trust. 

Also, it sounds like it's not well defined as to how to interpret the alignment of parameters between the method signature and the method descriptor. This will make it very hard in general to pair up parameter names with parameters, unless the arities all match perfectly (which they often don't in the case of Kotlin).

If one of you with a vested interest could at least report to the Kotlin bug tracker that Kotlin is not writing the same number of parameters into the `MethodParameters` block as contained in the method signature, I would appreciate it. Meanwhile, I'll try to find the best way to handle this brokenness gracefully.

PR #189, opened 2018-04-11
fixed some issues with dot graph generation where it would throw null…

seanrl:
…ptrs if you tried to enable field or method info.  Also modified the dot graph generation algorithm to generate UML-like nodes when field or method info are enabled

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -184,6 +184,12 @@
 		</dependency>
 
 		<dependency>
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId>
+			<version>24.1-jre</version>
+		</dependency>
+
+		<dependency>
 			<groupId>org.assertj</groupId>
 			<artifactId>assertj-core</artifactId>
 			<version>3.1.0</version>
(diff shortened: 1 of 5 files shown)


seanrl:
If users like this feature (I do), it could probably use some tweaks here and there.  The UML significantly increases the size of the graphs, so that can also be a problem.

lukehutch:
Thank you @seanrl, this is probably really useful. There are a few changes I'll need to make, e.g. I'm trying hard not to add any additional dependencies, so I'll have to add a custom HTML escape function. I probably have to put the UML generation behind a Boolean parameter or something. But I really appreciate the additions and the bugfixes!

lukehutch:
@seanrl I added a bunch more features (display of annotations, etc.), made some Graphviz formatting improvements, and I'm releasing the result as 2.18.2. Check out the output from that version. Thanks again!

lukehutch:
PS re. the UML increasing the size of graphs -- the way to fix that is to not call `.enableMethodInfo()` and/or `.enableFieldInfo()`. I may need to add more fine-grained controls if the graph sizes are problematic for people.

seanrl:
Hey @lukehutch thanks for accepting and expanding the capability.  I'll check out the new version soon.  Even with the expanded size of generated UML diagrams, I've had pretty good luck with some optional parameters in GraphViz's sfdp and neato programs.  Thanks for the info, and looking forward to checking out the annotations and such.

PR #198, opened 2018-06-04
Travis build: add 'oraclejdk9'

sullis:


lukehutch:
@sullis This test fails because JDK9 doesn't have an "rt.jar", and one of the tests tries scanning system classes by scanning this jar. There is no JDK9+ module path support yet, since there's no good way to scan the module path at this point (I'm trying to find a workaround). I don't think there's a need to bump the Travis JDK version to 9.

lukehutch:
No longer needed -- full JDK9 support is included with FCS version 3.0.0. Thanks for the PR.

PR #200, opened 2018-06-16
Update AssertJ to version 3.10.0

PascalSchumacher:
`StrictAssertions` has been replaced by `AssertionsForClassTypes`, but most of the time `Assertions` can be used instead of  `AssertionsForInterfaceTypes` or `AssertionsForClassTypes`.


lukehutch:
Thanks for this PR!

PR #204, opened 2018-06-24
Test on multiple JDKs

seanf:


seanf:
The build failed on all JDKs, but it turns out the build for `master` was already broken: https://travis-ci.org/lukehutch/fast-classpath-scanner/builds/395423627

lukehutch:
@seanf I had to disable JDK 8 testing, because of https://github.com/moditect/moditect/issues/62 .

However, the jars I build are built in JDK 7 compatibility mode. If you trust compatibility mode, hopefully this will ensure compatibility.

lukehutch:
Good to know. It looks like that would allow you to have multiple JDK versions installed. I have been tempted many times to try to get that working in Maven, but I think just using a source and binary level setting in a more recent JDK is a cleaner option. Thanks for the info though.

PR #211, opened 2018-07-17
Only use html5 arg for java9+.

johnou:
Add module info class on release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 jdk:
   # current LTS versions
-#  - openjdk8    # Disabled because moditect needs JDK 9+ (though we build in JDK 7 mode) 
+  - openjdk8
   # already EOL
 #  - openjdk9
   # feature release
--- a/pom.xml
+++ b/pom.xml
@@ -13,6 +13,8 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+		<argLine.javadocs>-Xdoclint:none</argLine.javadocs> <!-- Overridden in JDK9 -->
 	</properties>
 
 	<issueManagement>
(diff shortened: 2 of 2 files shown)


johnou:
@lukehutch looks like a unit test fails on java8.

lukehutch:
@johnou Thanks for this PR. The test failure has to do with the JDK 8 build pulling in a dep that the other builds don't pull in for some reason, so there are unexpected classes on the classpath. I'll fix that.

However, I'm curious as to why you pulled the Moditect plugin invocation into the release profile. How does that fix the issue with Moditect?

johnou:
@lukehutch it is a workaround to allow building the project with JDK 8 on CI.

johnou:
When I'm back from holiday I can assist with fixing the actual issue with Moditect.

lukehutch:
@johnou thanks. I won't check this in for now then, since I filed a bug report in the Moditect bug tracker, and I told the Moditect creator that he could just build FCS out of the box on JDK 7 or 8 to exhibit the issue. BTW there are still a few users on JDK 7, so that needs to work too.

johnou:
@lukehutch looks like a feature.. "If you derive the module info descriptor though (moduleInfo), then this indeed will only run on JDK 9 as this uses some JDK 9 APIs under the hood.", this essentially means if you want to keep the Moditect configuration as is moving it to the release profile is fine and requires releases are built with java9+. These changes should not prevent the library working with JRE 7.

lukehutch:
@johnou What's the source for that quote? The Moditect author said (in the bug I filed) that he considers Moditect not running under JDK 7 / 8 to be a bug.

johnou:
From the comment of the author in your bug report.

lukehutch:
Ah, I see, I didn't understand what the Moditect author meant by "(`moduleInfo`)". He was referring to a Maven element. Thanks for clarifying.

lukehutch:
I cherry-picked changes from your PR, since I decided to use the `modInfoFile` fix rather than the one in your PR. Thanks for the PR though, and thanks for the other Javadoc fixes! PS there's still an Eclipse (m2e) warning in the pom file, if you want to take a crack at that when you're free. :-)

PR #226, opened 2018-08-02
Close resources in finally block.

johnou:
Little cleanup.


johnou:
Better to _try_ and clean up if there is an IO failure, no need to rush a release for this PR.

lukehutch:
Yup, good changes, thanks.

PR #262, opened 2018-10-18
Add Diffblue tests

diffblue-assistant:
The test cases contained in this PR were generated by [Diffblue's](http://www.diffblue.com) [Deeptest](http://www.diffblue.com/papers/overview.pdf) software. 

One of our missions is to provide useful tests to open source projects. In return, we would appreciate your feedback about our tests to help us further improve our tool. 
    
Results for various open source projects can be viewed at:

   [http://diff.blue/demo](http://diff.blue/demo)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
 
 	<groupId>io.github.classgraph</groupId>
 	<artifactId>classgraph</artifactId>
-	<version>4.4.1-SNAPSHOT</version>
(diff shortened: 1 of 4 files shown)


lukehutch:
Thanks, but there is very little valuable that is tested in these automatically generated testcases. Also, since this PR is generated directly from the sourcecode, these tests really only test whether something *changes* in the source in future, not whether the current code is necessarily correct.

PR #268, opened 2018-10-26
GH-267: Make ClassGraph work with Spring Boot DevTools / RestartClassLoader.

### LINKED ISSUE #267: Class loading doesn't work reliable with Spring Boot's DevTools
Part of Spring Boot are the [Spring Boot Developer Tools](https://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-devtools.html). They support an "Automatic Restart" through the [Restart ClassLoader](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/devtools/restart/classloader/RestartClassLoader.html). 

From the docs:

> The restart technology provided by Spring Boot works by using two classloaders. Classes that do not change (for example, those from third-party jars) are loaded into a base classloader. Classes that you are actively developing are loaded into a restart classloader. When the application is restarted, the restart classloader is thrown away and a new one is created. 

We are using ClassGraph to find classes that should become [Neo4j OGM Entities](https://github.com/neo4j/neo4j-ogm), much like JPA / Hibernate entities. That works very well, in a normal class loader situation as well as in the RestartClassLoader.

However, as soon as we try to load classes found by ClassGraph we ran into issues with the _parent last_ approach of Springs Restart Classloader together with `ClassGraphClassLoader`. 

`ClassGraphClassLoader` does a superb job of finding the domain classes in the correct class loader but it loads them from the wrong one.

You'll find attached the verbose log of scanning with ClassGraph 4.4.8.

Basically what are we doing in OGM is this

```
	ScanResult scanResult =
		new ClassGraph()
			.verbose()
			.enableAllInfo()
			.whitelistPackages("com.example.some")
			.scan();

	Dingens getDingens() {
		try {
			return (Dingens) scanResult
				.getClassInfo("com.example.some.Dingens")
				.loadClass().getDeclaredConstructor().newInstance();
		} catch (InstantiationException | IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
			throw new RuntimeException(e);
		}
	}

	@Override
	public void run(String... args) throws Exception {
		System.out.println(getDingens());
	}
```

Dynamically loading classes, their methods and fields. The above method fails with a class cast exception because the directly referenced `Dingens` is part of `RestartClassLoader`, the dynamically loaded of `AppClassLoader`. This is despite the fact that the corresponding `ClassInfo` references `RestartClassLoader`. This is IMHO a bug.

We have a PR coming in for that problem. Attached you'll find the verbose log and the complete sample.

[verbose-log.txt](https://github.com/classgraph/classgraph/files/2518004/verbose-log.txt)
[classgraphplayground.zip](https://github.com/classgraph/classgraph/files/2518006/classgraphplayground.zip)

michael-simons:
This commit has to changes:

- It introduces SpringBootRestartClassLoaderHandler and adds it to the registry. RestartClassLoader is parent last and such should be explicitly configured.
- The codewise small but impact wise bigger change is instantiating ClassGraphClassLoader with a call of super(null), eliminating it’s parent.

The later one is done due to the following reason:
ClassGraphClassLoader goes through lengths in `io.github.classgraph.ClassGraphClassLoader#findClass` to determine which class loader has been the source of the scanned class and that works very well.

`findClass` is however called from `loadClass` only if there is no parent class loader or the parent class loader doesn't find the class.

In Spring Boot together with Spring Boots Restart Class Loader we have the following situation:

A launcher thread spawns the actually application using the `RestartClassLoader` described in #267. 

From there on, all application classes (especially entities from OGM) live in the RestartClassLoader. 

ClassGraph finds them as well and also notices that they are part of RestartClassLoader.

however, `loadClass` always uses the parent class loader first if any and none of the class loaders that are determined in `findClass`. 

The main change here is not giving `ClassGraphClassLoader` a parent, thus forcing it to do the lookup by himself with the mechanism of findClass.

The rest are just tests.

* Create a Thread using another class loader (knowing only 2 classes)
* Run everything from that Thread and Class loader
* Create a test stub that is runnable from the above thread and also in a normal way (Boot devtools on and off)
* See if scanned classes are compatible with directly loaded classes.

If you have any questions or suggest that it would be better to solve this by also overwriting `loadClass`, I'm very happy to discuss.

This closes #267.

### DIFF
--- a/src/main/java/io/github/classgraph/ClassGraphClassLoader.java
+++ b/src/main/java/io/github/classgraph/ClassGraphClassLoader.java
@@ -45,7 +45,7 @@ class ClassGraphClassLoader extends ClassLoader {
      *            The ScanResult.
      */
     ClassGraphClassLoader(final ScanResult scanResult) {
-        super();
+        super(null);
         this.scanResult = scanResult;
         registerAsParallelCapable();
     }
--- a/src/main/java/io/github/classgraph/classloaderhandler/ClassLoaderHandlerRegistry.java
+++ b/src/main/java/io/github/classgraph/classloaderhandler/ClassLoaderHandlerRegistry.java
@@ -53,6 +53,7 @@ public class ClassLoaderHandlerRegistry {
                 new ClassLoaderHandlerRegistryEntry(WebsphereLibertyClassLoaderHandler.class),
                 new ClassLoaderHandlerRegistryEntry(WebsphereTraditionalClassLoaderHandler.class),
                 new ClassLoaderHandlerRegistryEntry(OSGiDefaultClassLoaderHandler.class),
+                new ClassLoaderHandlerRegistryEntry(SpringBootRestartClassLoaderHandler.class),
 
                 // JPMS support
                 new ClassLoaderHandlerRegistryEntry(JPMSClassLoaderHandler.class),
--- a/src/main/java/io/github/classgraph/classloaderhandler/SpringBootRestartClassLoaderHandler.java
+++ b/src/main/java/io/github/classgraph/classloaderhandler/SpringBootRestartClassLoaderHandler.java
@@ -1,0 +1,69 @@
+/*
+ * This file is part of ClassGraph.
+ *
+ * Author: Michael J. Simons
+ *
+ * Hosted at: https://github.com/classgraph/classgraph
+ *
+ * --
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2018 Luke Hutchison
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+ * documentation files (the "Software"), to deal in the Software without restriction, including without
+ * limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so, subject to the following
+ * conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all copies or substantial
+ * portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT
+ * LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO
+ * EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN
+ * AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE
+ * OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package io.github.classgraph.classloaderhandler;
+
+import io.github.classgraph.ClassLoaderHandler;
+import io.github.classgraph.ScanSpec;
+import io.github.classgraph.utils.ClasspathOrder;
+import io.github.classgraph.utils.LogNode;
+
+/**
+ * This handler uses {@link io.github.classgraph.ClassLoaderHandler.DelegationOrder#PARENT_LAST} to support
+ * the <code>RestartClassLoader</code> of Spring Boot's devtools. <code>RestartClassLoader</code> provides
+ * parent last loading for specified URLs (those are all that are supposed to be changed during development).
+ * Therefor the handler for that class loader also has to delegate in <code>PARENT_LAST</code> order.
+ */
+class SpringBootRestartClassLoaderHandler implements ClassLoaderHandler {
+
+	// Spring Boots devtools class loader is an extension of URLClassLoader so there's no need
+	// to use reflection to access the urls supported and we can delegate the handling to an
+	// internal instance of URLClassLoaderHandler.
+	private final URLClassLoaderHandler handlerDelegate = new URLClassLoaderHandler();
+
+	@Override
+	public String[] handledClassLoaders() {
+		return new String[] { //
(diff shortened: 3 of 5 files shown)


lukehutch:
@michael-simons this is incredible -- I don't think I would have ever found the need to call `super(null)` to fix this -- and more importantly, your PR is (I think) the biggest and most complex PR I have ever received for ClassGraph! I really want to thank you for your contribution!

I'll pull in as-is, but I have one question -- is checking if the `getSimpleName()` of the initial context ClassLoader is `"AppClassLoader"` robust and future-proof? (The entire classloader hierarchy changed in JDK9, for example...)

lukehutch:
PS @michael-simons I really could have used your help on ClassGraph a long time ago! :-) Supporting Spring Boot has been a major pain point for the project before now! (It all seems to be working OK now, but it took a lot of work to get to that point...)

lukehutch:
@michael-simons Actually I have another question -- why do you delegate from `SpringBootRestartClassLoaderHandler` to `URLClassLoaderHandler`? `URLClassLoaderHandler` is supposed to handle all subclasses of `URLClassLoader` already. Was that not working?`

https://github.com/classgraph/classgraph/blob/master/src/main/java/io/github/classgraph/utils/ClasspathFinder.java#L133

michael-simons:
I think that the handler as I understood it should use the same semantics in looking for classes than the class loader itself. `RestartClassLoader` does `PARENT_LAST`. If there's another way to specify it, the better.

lukehutch:
Oh, good point, I saw that in the code and then promptly forgot about that difference :-)  And you even explained this in the first line of your PR comment! So that it's clear for future reference, what could be added in a comment explaining why this classloader is `PARENT_LAST`? What exactly goes wrong if that is `PARENT_FIRST`? Maybe all subclasses of `URLClassLoader` should be handled `PARENT_LAST`, since they are overriding the default mechanism? And actually, this is the first time `PARENT_LAST` has ever been tested, so I'm glad it works!

michael-simons:
Something along the line

"Spring Boot's RestartClassLoader sits in front of the parent class loader and watches a given set of directories for changes. While those classes are reachable from the parent class loader directly, they should always be loaded through direct access from the RestartClassLoader until it's completely turned of by means of Spring Boot Developer tools."

The RestartClassLoader shades only the project classes and additional directories that are configurable, so itself needs access to parent, but last.

lukehutch:
Sorry, ignore that last comment, I somehow missed the fact that you answered all my questions in #267. I'll put a link to that in the `PARENT_LAST` line of new `ClassLoaderHandler`.

lukehutch:
Great, I'll add that, thanks!

GedMarc:
This is very interesting for me - 

I wonder how with the modular system this would take effect - I've had to implement auto restarts with SPI and on clearing and updating I get the results correctly from ClassGraph, but the modular reloading I haven't found a solution to it yet,

I've been a bit worried about Spring and the modular system, FAT jar's aren't exactly JPMS compatible at all, and the classloader as far as I'm aware doesn't really play a big role anymore (outside of classpath mode obviously)

Wonder what the impact is going to be for me

lukehutch:
Released in 4.4.9: https://github.com/classgraph/classgraph/releases/tag/classgraph-4.4.9

Thanks again for your contributions, @michael-simons!

Please let me know whether you think all subclasses of `URLClassLoader` should default to `PARENT_LAST`. I'm having a hard time getting my head around whether that would be a sensible default, because (despite working on ClassGraph for three years now) I still don't know enough about all the different ClassLoaders that are found out there in the wild.

michael-simons:
Fantastic. I'll reach out to the Pivotal people about the order.

lukehutch:
@GedMarc I was wondering about this recently too -- specifically the part "the classloader as far as I'm aware doesn't really play a big role anymore (outside of classpath mode obviously)". I think that actually what we'll see happening in the future is a lot of these large runtime environments will modify their classloaders to define their own `ModuleLayer` and associated configuration, i.e. they will actually upgrade themselves to work more like the built in JPMS classloaders.

lukehutch:
@michael-simons I made a ton of changes to ClassGraph recently, and in the process managed to break your test, and I'm not sure how :-S  (`io.github.classgraph.issues.issue267`)

The actual test passes, called from `ClassLoadingWorksWithParentLastLoadersStub::sameClassLoaderThatFoundAClassShouldLoadIt()`. However, once `TestLauncher::run()` launches, there is an assertion failure (it doesn't cause a JUnit failure, because the failure is in a separate launched thread):

```
Exception in thread "Thread-1" java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at io.github.classgraph.issues.issue267.TestLauncher.run(ClassLoadingWorksWithParentLastLoadersStub.java:71)
Caused by: org.junit.ComparisonFailure: expected:<"[FakeRestart]ClassLoader"> but was:<"[App]ClassLoader">
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at io.github.classgraph.issues.issue267.ClassLoadingWorksWithParentLastLoaders.assertCorrectClassLoaders(ClassLoadingWorksWithParentLastLoaders.java:72)
	... 5 more
```

I would appreciate your help looking into this, since I'm not sure what's wrong...

lukehutch:
OK, I figured it out -- `FakeRestartClassLoader` needed to have its own `ClassLoaderHandler` that specified `PARENT_LAST` delegation mode, otherwise the parent classloader (`AppClassLoader`) would resolve the class (since the default delegation mode is `PARENT_FIRST`). I updated the test in the above commit.

PR #269, opened 2018-10-26
Fix hardcoded name of parent class loader in test.

michael-simons:
This makes the test more future proof and robust.


michael-simons:
Oh btw, I happily apply any code formatting rules you might have so that you don't have to polish stuff afterwards.

lukehutch:
Thanks for the fix!

PR #270, opened 2018-10-26
Add getDeclaredAnnotationInfo() and getDeclaredAnnotationInfo(String).

michael-simons:
We would love to see a corresponding `getDeclaredAnnotationInfo` method that works the same way as with fields and methods (they already come in pairs as well). This PR adds this feature.

### DIFF
--- a/src/main/java/io/github/classgraph/ClassInfo.java
+++ b/src/main/java/io/github/classgraph/ClassInfo.java
@@ -1211,6 +1211,37 @@ public class ClassInfo extends ScanResultObject implements Comparable<ClassInfo>
     // Annotations
 
     /**
+     * Returns information on visible methods declared by this class, but not by its interfaces or superclasses,
+     *
+     * <p>
+     * It does not include any the {@link Inherited} meta-annotation.
+     *
+     * @return A list of {@link AnnotationInfo} objects for the annotations on this class, or the empty list if
+     *         none.
+     */
+    public AnnotationInfoList getDeclaredAnnotationInfo() {
+        if (!scanResult.scanSpec.enableAnnotationInfo) {
+            throw new IllegalArgumentException("Please call ClassGraph#enableAnnotationInfo() before #scan()");
+        }
+        return annotationInfo == null ? AnnotationInfoList.EMPTY_LIST : annotationInfo;
+    }
+
+    /**
+     * Get a the named annotation on this class, or null if the class does not have the named annotation.
+     *
+     * <p>
+     * It does not include any the {@link Inherited} meta-annotation.
+     *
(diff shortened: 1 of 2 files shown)


lukehutch:
A few things --

* The `getDeclaredAnnotationInfo(final String annotationName)` method should call `getDeclaredAnnotationInfo()`, not `getAnnotationInfo()`.
* The `getDeclaredAnnotationInfo()` Javadoc comment needs updating, it still refers to methods :-)  And actually if the two things listed in the previous bullet point are the intent, the comments on these two methods should reflect how they are different from the other `getAnnotationInfo()` methods. Also the other `getAnnotationInfo()` methods should reference the `getDeclaredAnnotationInfo()` methods.
* I think this might be better named `getDirectAnnotationInfo()`? I think the intent is that (1) annotations should not be inherited from superclasses or implemented interfaces if the annotation has the `@Inherited` meta-annotation, and (2) that for any annotations on the class, meta-annotations should not be returned, i.e. only direct annotations on the class should be returned. This is quite a different intent or semantics than `getDeclaredMethodInfo()` and `getDeclaredFieldInfo()`. The semantics are actually a better fit for what is returned by the `ClassInfoList#getDirect()` method, hence the suggested different name.

michael-simons:
First comment, first bullet: I was thinking about that. Don’t know why didn’t do it in the first place. 

And regarding the semantics: I’ll need a moment to think about your intent, but it sounds reasonable. 
I keep that PR open and will force push any further changes to my branch. 

Thank you for your input!

lukehutch:
@michael-simons the change I'm suggesting is bigger than you were probably expecting to implement, so I totally understand if you would rather not make the change and instead you just wish the feature existed. I can certainly make the change, it just depends how involved you want to get with ClassGraph development. Let me know whether you want to take a shot or not. One possibility is that I could write the code and you could write the unit test to make sure the code is doing what you need.

michael-simons:
Thanks for the warning. 
As I see it the difficulty is not extending `AnnotationInfoList` but determination of what annotation is direct or not, especially for methods. 

Anyway, as we are betting to a large extend on ClassGraph, I don’t have a problem with investing time into this. A unit test is the least we would provide. 
As long as no one else request this, don’t feel urged to implement it yourself. I’ll have a more detailed look tonight or next week.

lukehutch:
Really it shouldn't be too difficult if you follow the pattern in `ClassInfoList`. It stores the directly related classes in an internal set, which is read by the filtering methods and `directOnly()`. In this case the direct annotations passed into that set via the `AnnotationInfoList` constructor simply won't include the `@Inherited` annotations or the meta-annotations.

PR #271, opened 2018-10-27
Fix retrieval of inherited annotations classes.

michael-simons:
While working on the support for `directOnly()` for annotation lists, I found this subtle, little bug.


lukehutch:
Good catch! I don't think I ever fully tested this :-)

PR #272, opened 2018-10-27
Add directOnly() to public API of AnnotationInfoList.

michael-simons:
This allows to filter the list of annotations to direct annotations only. Those are
all annotations directly declared on a element.

The only case directOnly can deliver something else are classes that inherited classes
annotated with meta-annotated @Inherited annotations.

*Note* This PR is rebased onto my branch `fix-retrieval-of-annotation-classes` and should be merged after #271 is merged.

Closes #270 in favor of this PR.

### DIFF
--- a/src/main/java/io/github/classgraph/AnnotationInfoList.java
+++ b/src/main/java/io/github/classgraph/AnnotationInfoList.java
@@ -29,21 +29,41 @@
 package io.github.classgraph;
 
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 
 import io.github.classgraph.InfoList.MappableInfoList;
 
 /** A list of {@link AnnotationInfo} objects. */
 public class AnnotationInfoList extends MappableInfoList<AnnotationInfo> {
+
+    /**
+     * The set of annotations directly related to a class or method and not inherited through a meta-annotated annotation.
+     * This field is nullable, as they annotation info list is build incremental. To fill it accordingly, we would have
+     * to overwrite all list methods. Instead, checks are handled in {@link #directOnly()}.
+     */
+    private final AnnotationInfoList directlyRelatedAnnotations;
+
     AnnotationInfoList() {
         super();
+        this.directlyRelatedAnnotations = null;
     }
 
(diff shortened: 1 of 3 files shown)


lukehutch:
@michael-simons Nice work! Thanks so much for implementing this. It's a good addition to the API.

lukehutch:
@michael-simons OK, so now that I'm looking at this more closely, I realize I was wrong in thinking that all `AnnotationInfoList` instances may need to have `.directOnly()` called on them, because actually it is only class annotations (not method, method parameter, or field annotations) that have annotations inherited with `@Inherited`. So maybe your original idea of adding methods `ClassInfo#getDirectAnnotationInfo()` would have made more sense. You've done the work now, but what do you think is better?

I agree with your comment that it's a bit weird how the direct annotations are constructed (the original `AnnotationInfoList` is built incrementally, then when you call `ClassInfo#getAnnotationInfo()`, it builds a new list, adding in inherited annotations).

Also, Another thing I just realized is that the code as it is currently implemented does not pull in meta-annotations in `ClassInfo#getAnnotationInfo()` (since the `AnnotationInfo` objects are just the direct annotations pulled in when the classfile is read), but `ClassInfo#getAnnotations()` does pull in meta-annotations. Presumably meta-annotations would not be returned if you call `.directOnly()`, but that's another thing that will need to be skipped, along with `@Inherited`.

lukehutch:
Actually I take that back -- there can still be meta-annotations on fields, methods, etc., so `.directOnly()` probably does make sense :-)  I'll get meta-annotations working.

lukehutch:
Just committed a change. I didn't realize meta-annotation support wasn't actually working since the changeover from FastClasspathScanner to ClassGraph. I think that the committed change has the right semantics, but please check the unit test changes to see if it's what you would expect. Thanks!

michael-simons:
Sweet. Exactly what (at least) I would expect. Thanks.

lukehutch:
Released in 4.4.10. Thanks again for your contributions, @michael-simons !

https://github.com/classgraph/classgraph/releases/tag/classgraph-4.4.10

Issue #273, opened 2018-10-28
SPI Config JPMS


GedMarc:
Caused by: java.util.ServiceConfigurationError: io.github.classgraph.ClassLoaderHandler: module io.github.classgraph does not declare `uses`

JDK 10 and 11 for your SPI's, can you place the uses clause in your module.info :)

GedMarc:
Caused by: java.util.ServiceConfigurationError: io.github.classgraph.ClassLoaderHandler: module io.github.classgraph does not declare `uses`
	at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:588)
	at java.base/java.util.ServiceLoader.checkCaller(ServiceLoader.java:574)
	at java.base/java.util.ServiceLoader.<init>(ServiceLoader.java:503)
	at java.base/java.util.ServiceLoader.load(ServiceLoader.java:1691)
	at io.github.classgraph@4.4.10/io.github.classgraph.classloaderhandler.ClassLoaderHandlerRegistry.loadAdditionalClassLoaderHandlers(ClassLoaderHandlerRegistry.java:137)
	at io.github.classgraph@4.4.10/io.github.classgraph.classloaderhandler.ClassLoaderHandlerRegistry.<clinit>(ClassLoaderHandlerRegistry.java:67)
	at io.github.classgraph@4.4.10/io.github.classgraph.utils.ClasspathFinder.<init>(ClasspathFinder.java:229)
	at io.github.classgraph@4.4.10/io.github.classgraph.Scanner.call(Scanner.java:411)

lukehutch:
@michael-simons I think this is for the service discovery you added / got working... Sounds like a one-line fix...

GedMarc:
![image](https://user-images.githubusercontent.com/5367513/47618745-70cf5c00-dadf-11e8-9017-ebbc5a95c94b.png)
 should do it

lukehutch:
@GedMarc Thanks for the info, I didn't know about that. @michael-simons the descriptor is in `src/moditect`. But actually I'll need to push out a new release anyway, so since it's such a simple change, I can save you the trouble of sending me a PR -- I'll fix it. Sorry to bother you!

lukehutch:
@GedMarc Interesting that this doesn't bomb until runtime when you try to actually use the service. Runtime failures that could be statically detected (e.g. at the time you try to load the module) are an antipattern :-/

michael-simons:
[...]

PR #274, opened 2018-10-28
Export utils as well (#273).

michael-simons:
You have to export the utils package as well, otherwise the calling site on the module path cannot implement a custom class loader handler. See attached example.

[classgraphtest-issue273.zip](https://github.com/classgraph/classgraph/files/2522966/classgraphtest-issue273.zip)

### DIFF
--- a/src/moditect/module-info.java
+++ b/src/moditect/module-info.java
@@ -1,4 +1,6 @@
 module io.github.classgraph {
     exports io.github.classgraph;
+    exports io.github.classgraph.utils;
+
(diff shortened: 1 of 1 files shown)


lukehutch:
I really don't want to export `utils` (see #273). Closing (thanks anyway!).

michael-simons:
Having had a deeper. look into this package, it’s fully understandable. +1

lukehutch:
Yeah, I already don't like how many classes I had to put in the main package just to expose as few classes as possible as public. This would make it significantly worse :-S

PR #280, opened 2018-11-03
Add Unit test to measure the performance of nested jar scans

larsgrefer:


lukehutch:
See other content re. putting performance tests here, and using JMH:

https://github.com/classgraph/classgraph/tree/master/src/test/perf/io/github/classgraph

larsgrefer:
I moved the test. How can I execute it now? there seems to be no maven goal for this.

larsgrefer:
Closing this, because #281 is closed

PR #281, opened 2018-11-03
Use the spring-boot-loader JarFile implementation when possible to prevent extracting nested jars

larsgrefer:
Based on a discussion in https://gitter.im/classgraph/Lobby

This implementation seems to be slightly slower than the current master branch but I open this PR anyway, so it's easier to discuss this patch.


lukehutch:
I'll hold off on merging this until it's reflection-based, and until we get to the bottom of the performance discrepancy.

larsgrefer:
Why should it use reflection?

lukehutch:
It sounds like you're thinking this is not worth it due to the performance impact. If you do make a final determination that this is the case, go ahead and close this PR (and the associated perf testing PR), although if you do figure out how to make this faster, we can always reopen it.

larsgrefer:
Currently, this PR has no benefit, because its slower than the current implementation, therefore it should not be merged.

lukehutch:
OK -- thanks for giving this a shot! It was worth trying out this approach.

larsgrefer:
Yeah, it was fun.

lukehutch:
Well you attacked the problem in the middle of one of the most complex parts of ClassGraph -- I'm glad there are programmers out there that find "tactical ops" like that fun! ;-)

larsgrefer:
The performance difference seems to be in the constructor of `io.github.classgraph.utils.JarfileMetadataReader` while iterating over the zip entries.

The for loop takes 50 to 70ms to execute using the spring boot jar file while the unzipped one takes >10ms. This makes the spring boot implementation so much slower, that it's more efficient to extract the nested jar first.

lukehutch:
> The performance difference seems to be in the constructor of `io.github.classgraph.utils.JarfileMetadataReader` while iterating over the zip entries.

That's pretty interesting. This does sound like the difference between `ZipFile` (which only reads the zip central directory) and `ZipInputStream` (which iterates through all entries, and is forced to unzip or extract each entry, not just the metadata).

Since they have a zipfile central directory parser, there's no real reason why their `JarFileEntries` should be much slower than ClassGraph's approach unless they were doing unnecessary work to unzip or extract every entry. ClassGraph has to extract (assuming stored, not compressed) all nested jars, but does not have to uncompress the contents of the leaf-node jars, it only iterates through the entries via the `ZipFile` API. (`ZipFile` can only work on whole files though, as described previously.) Therefore, if Spring Boot is slower, and it's not even extracting nested jars to files, it must be doing unnecessary decompression/extraction work that ClassGraph is not doing, just to read the nested jar directory. (This is assuming that the central directory metadata parsing code in Spring Boot is not dramatically different in speed than the `ZipFile` code, which it should not be.)

It is possible that the Spring Boot code only uses their central directory parser for the toplevel jar, and doesn't use it for the nested (stored) jar. If that is the case, you could solve the performance problem by fixing their code to reuse their central directory parser for nested (stored) jars, and making `JarFileEntries` only fetch and decompress the jar entry contents lazily / on demand.

The only downside to doing this should be that the size of an entry will be only an estimate until the contents of the entry is decompressed, and then it can be known with certainty. (Zipfile directory entries contain an uncompressed size, but sometimes it is 0 or -1, and sometimes it is nonzero positive, but wrong, so it is only a size hint.)

larsgrefer:
I've tracked the performance difference down to the
```
final ZipEntry zipEntry = iter.nextElement();
```
call in the beforementioned `for`-loop in `JarfileMetadataReader`

This call ends up in `org.springframework.boot.loader.jar.JarFileEntries#getEntry(int, java.lang.Class<T>, boolean)`

lukehutch:
`nextElement()` is really fast for `ZipFile`...

larsgrefer:
The repeated calls to `nextElement()` are causing the underlying `RandomAccessFile` to jump back and forth in order to re-read the header entries from the central directory.

The entries aren't cached since https://github.com/spring-projects/spring-boot/commit/e2368b909b46bc5bcec6792fb208ba9bd0fe6aaa

larsgrefer:
The _Spring Boot_ implementation reads the central directory sequentially when the `JarFile` is created, but it does not store the complete entries in memory but only their hashes and their positions in order to reduce the memory consumption.

As explained in the commit message of https://github.com/spring-projects/spring-boot/commit/e2368b909b46bc5bcec6792fb208ba9bd0fe6aaa this behavior is sufficient for the use case of classloading, but it seems inefficient for Iterating over the entries

larsgrefer:
Based on this information, I might be able to implement a less memory efficient but therefore faster ZipFile class

GedMarc:
Please no dependencies, and most certainly no framework dependencies!

Especially for those of us that have no interest in using spring boot! Pulling in performance updates 100%, adding in spring boot, hell no, PLEASE PLEASE!

lukehutch:
@GedMarc don't worry, I am 100% committed to not adding runtime dependencies. Call me crazy, but I'm so committed to not adding dependencies that I wrote a JSON serializer and deserializer from scratch for ClassGraph, to avoid depending upon a library. (Do you have any idea how hard it is to write a Java serializer and deserializer from scratch, also handling object graph cycles and generic types? Never again :-p )  This is why I told Lars that any calls to Spring APIs need to use reflection.

larsgrefer:
@GedMarc The "_dependency_" on `spring-boot-loader` would be fully optional. The plan was to detect if the classes are present, and if so, use them to make ClassGraph faster. Using this approach ClassGraph would automatically faster for people which are using _Spring Boot_ anyway. (And it would be possible to just throw in the 90KB `spring-boot-loader.jar` whithout actually using _Spring Boot_ just to make ClassGraph faster for nested jars)

@lukehutch I think overriding `org.springframework.boot.loader.jar.JarFileEntries#visitFileHeader` and `org.springframework.boot.loader.jar.JarFileEntries#add` in order to store the full header and not just it's position and the hashCode of the name would prevent `org.springframework.boot.loader.jar.JarFileEntries#getEntry(int, java.lang.Class<T>, boolean)` from re-reading the header, thus making the iteration way faster. Unfortunately `JarFileEntries` is package-private and the beforementioned methods are private, so at least some of the classes would need to be re-implemented in ClassGraph.

lukehutch:
@larsgrefer I suggest submitting a patch to Spring-Boot that adds a configuration API call that causes all `ZipEntry` values to be cached, not just the hashes (or maybe a call that will produce an enumeration or iterator that loads all entries at the start of the enumeration or iteration).

lukehutch:
Another possibility is for ClassGraph to manually parse `ZipEntry` objects from jarfile central directories. The `ZipEntry` class is not final, and the fields are not private. Therefore, the work done by the `ZipFile` constructor could be replicated in `ClassGraph`. For non-nested zipfiles, the `ZipEntry` could be passed into a `ZipFile` instance to extract resources. For a nested jar, the `java.util.zip.Inflater` class might be able to be used to directly extract deflated zip entries without writing custom decompression code. (I have no idea if zip entry contents are a simple deflated stream, or if there is some custom handling that is required to unzip files.) Also there are a range of codecs that would probably need to be supported. I'll have to look at the JDK `ZipFile` code and see if this is a possibility.

(I'm traveling for the next week, so progress on my end will be slow)

larsgrefer:
I've opened spring-projects/spring-boot#15096 to see what the spring guys think about this.

lukehutch:
@larsgrefer I took a closer look at this. I thought that `ZipEntry` contained a file offset for the start of the data for the entry, but actually it just contains a filename and size, and the start offset is obtained from a hash map stored in the `ZipFile` the entry was obtained from. (I was going to try to manually parse out `ZipEntry` values for nested jars, and then spoof the data offset for the zip entry to work with the nested jar rather than the outer jar.) There is no way to create the hash map in the outer `ZipFile` from anything other than the central directory of the outer jar.

The `ZipFile` code seems unnecessarily complicated though, and it might actually be worth it to write a simpler custom replacement for both `ZipFile` and `ZipEntry`. I would really like to avoid creating temporary files anyway, and this would achieve that.

It will be quite a bit of work, but I'll try to give it a shot in the near future.

lukehutch:
@larsgrefer I see you're still working on this -- did you conclude that if a build-time scan is employed, the overhead of calling the Spring Boot API (rather than using ClassGraph's internal API) is worth it?

larsgrefer:
@lukehutch The current overhead of the spring boot JarFile isn't worth it. I want to make things faster, not slower.

I'm currently working on two different approaches, to make the startup time of (repacked) JoinFaces-based _Spring Boot_ applications faster:

- I'm working on spring-projects/spring-boot#15096 which should remove the overhead. If that issue gets resolved, this PR here should make ClassGraph noticeably faster instead of slower (Thats why this PR is still open)
- I'm also working on a build-time scan solution, so the classpath scan at runtime can be avoided completely. The build-time scan happens before the jar file gets repackaged, so ClassGraph can use it's default `ZipFile`-based implementation here.

lukehutch:
Ah, that makes sense. Thanks.

I'm still interested in implementing a custom zipfile central directory parser, so that non-Spring Boot projects get similar benefits. Just got back from a couple of weeks of traveling, so I'll try to work on this soon.

lukehutch:
@larsgrefer ClassGraph now implements a custom zipfile central directory parser, so this PR is no longer needed (but thanks again for giving this a shot). I'll close the PR, but please test the master version, and let me know what sort of performance improvement you see, as well as any behavior regressions.

(The performance improvements will come from not having to extract the inner jarfiles to temporary files on disk just to read the paths of all entries in (stored, not deflated) nested jarfiles. However, if you are still scanning all classfiles in every jarfile, rather than a specific narrow set of packages, then the scan time will still be dominated by classfile scanning, not by nested zipfile extraction.)

PR #292, opened 2018-12-18
Add support for newer versions of equinox for classpath resolving

zoliMolnar:
Hey @lukehutch

The built in class loader handlers work fine in Eclipse up until Photon, but not when you try to use ClassGraph in Eclipse 2018-09 or newer.
This version changes the name of the Java field `cp` to `nestedDirName` in class `NestedDirBundleFile` [see commit](https://github.com/eclipse/rt.equinox.framework/commit/847091d8de7ec31db33fa4d3d7189eb872330eba#diff-846a7057d129eb1904817dd1af240fa7L36) so `EquinoxClassLoaderHandler` is unable to recognize the class path entry properly and appends the output directory as path prefix resulting in ignored class files during scanning.

The pull request aims to fix this issue while maintaining support for older versions.

Kind regards,
Zoltán Molnár


lukehutch:
Released in 4.6.9.

zoliMolnar:
Thank you for the quick response and release!

PR #293, opened 2018-12-19
Fix field name in EquinoxClassLoaderHandler to nestedDirName

zoliMolnar:
Hey @lukehutch I'm terribly sorry about it, but I've made a mistake on one of the field names that I've added yesterday by accidentally copy pasting an incorrect field name. My mistake doesn't cause any issues with previous versions of equinox class path resolving. 

This pull request aims to fix the issue above. I hope I haven't ruined anything and thank you for responding and releasing my fix yesterday so fast.

Kind Regards,
Zoltán Molnár


lukehutch:
@zoliMolnar Nothing at all ruined, and sorry I moved a bit too fast yesterday! :-)  You're the first person I know of who has even tried to use this with the latest Equinox, so I don't think this affects anyone else yet. I always try to push out a new release quickly anytime there's a bug that causes breakage, so I'll just do that again now. Thanks for the fix!

lukehutch:
Released in 4.6.10.

zoliMolnar:
Thank you for your fast response yet again! :)

Issue #294, opened 2018-12-28
Support for java - loggers


gitmotte:
I'd like to add my own PrintStream/Logger for logging. The LogNode - class does not support that and logs per default to System.err.

The standard - logging system of java is java.util.logging.Logger, there are many bridges for other logging-facades and system, therefore that would be enough flexible and does not add any dependencies.

lukehutch:
The reason for the LogNode class is that ClassGraph is aggressively parallelized, and mixing together log output from different threads can produce a jumbled order of log entries that is very hard to use for debugging. LogNode also arranges the output into a nice hierarchy for understandability.

LogNode does have an "immediate mode" that writes out log lines as they are generated, and you could add an option to this to write to a Java logger rather than to stderr (if you want to submit a PR) -- and/or you could redirect the hierarchical log output generated at the end to a Java logger.

Unfortunately I never managed to get my head around how all the different Java logging frameworks were supposed to work together, so I wouldnt even know how to get started with this. 

I do know though that Java loggers can capture stderr output, so that would give you a quick workaround, without having to write any code.

LogNode is pretty bare-bones right now, so feel free to improve it!

gitmotte:
[scan.log](https://github.com/classgraph/classgraph/files/2715738/scan.log)

gitmotte:
Thanks for merging!

lukehutch:
@gitmotte I noticed that if any of the ClassGraph tests that run before `LogNodeTest` enable verbose logging by calling `.verbose()` before `.scan()`, then when `LogNodeTest` runs, its test fails. Specifically `systemErrMessages` ends up empty:

https://github.com/classgraph/classgraph/blob/master/src/test/java/io/github/classgraph/test/utils/LogNodeTest.java#L31

Do you know how to fix this, to isolate the logging in this test?

lukehutch:
(reopening)

lukehutch:
Many thanks for the fix!

PR #295, opened 2018-12-28
#294 Support for java - loggers

gitmotte:
LogLevel INFO is by default enabled, including a default ConsoleLogger
(prints to System.err)

https://github.com/classgraph/classgraph/issues/294

### DIFF
--- a/src/main/java/nonapi/io/github/classgraph/utils/LogNode.java
+++ b/src/main/java/nonapi/io/github/classgraph/utils/LogNode.java
@@ -38,6 +38,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.logging.Logger;
 
 import io.github.classgraph.ClassGraph;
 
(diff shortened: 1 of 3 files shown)


lukehutch:
What happens if `logging.properties` is included in `main` rather than `test`? Will that serve as the default logging config file unless a user provides their own? Or could it override a user's `logging.properties` file? If it is not included in `main`, will a user have to provide `logging.properties` to get any sort of log output?

ClassGraph adds a timestamp and a label "ClassGraph" to each log entry, which should not be needed if this is sent to the standard Java logger. You can remove those parts of the log lines.

gitmotte:
The logging.properties should be located in src/test because it's just for the unit-test. If a user does not provide a logging.properties the default logging.properties of jre (located at JRE_DIRECTORY/jre/lib/logging.properties 
) will be the default. (configured with INFO, ConsoleHandler)
If the user has another logging-facade like me, the logging-configuration-file will differ (log4j2.xml or others) 
The logging will not occur unless ClassGraph.verbose() or ClassGraph.enableRealtimeLogging() will be called. 
Yes, the label "ClassGraph" may not be needed any more, but the LogNode collects multiple messages  on multiple lines and logs at the end of scanning. The time may differ from the logging-time, but the label may be duplicated 

example of logging-message configured with log4j2.xml

> 2018-12-29 00:58:08,125  INFO [ClassGraph-worker-0] [i.g.c.ClassGraph#flush] - 2018-12-29T00:58:07.733+0100	ClassGraph	ClassGraph version 4.6.11-SNAPSHOT

PR #299, opened 2019-01-15
Support for java - loggers #294

gitmotte:
fix for:  https://github.com/classgraph/classgraph/issues/294#issuecomment-454226138

"if any of the ClassGraph tests that run before LogNodeTest enable
verbose logging by calling .verbose() before .scan(), then when
LogNodeTest runs, its test fails. Specifically systemErrMessages ends up
empty"

> add a ConsoleHandler after setting System.setErr specific for this
testcase and remove it afterwards.

> remove specific logging.properties - default logging.properties of jre
will be enough.


gitmotte:
The command "~/bin/install-jdk.sh --target "/home/travis/openjdk11" --workspace "/home/travis/.cache/install-jdk" --feature "11" --license "GPL" --cacerts" failed and exited with 7 during .

don't know why this fails, @lukehutch do you have an idea?

gitmotte:
Did a openjdk11 build on windows with success.

lukehutch:
> don't know why this fails, @lukehutch do you have an idea?

I see these failures in Travis all the time. They are transient failures. I don't know what the issue is, but it's not a problem with the build itself.

lukehutch:
PS many thanks for submitting a solution right away! See question above about System.setErr, but otherwise, looks good to me.

lukehutch:
@gitmotte Awesome, thanks so much for these fixes! Much appreciated.

gitmotte:
:) thank you, too, for this great library!

PR #302, opened 2019-01-18
add classpath entries from fragments

eddi-weiss:


src/main/java/nonapi/io/github/classgraph/classloaderhandler/EquinoxClassLoaderHandler.java
```diff
@@ -102,15 +102,15 @@ public void handle(final ScanSpec scanSpec, final ClassLoader classLoader,
             final ClasspathOrder classpathOrderOut, final LogNode log) {
         // type ClasspathManager
         final Object manager = ReflectionUtils.getFieldVal(classLoader, "manager", false);
-        // type ClasspathEntry[]
-        final Object entries = ReflectionUtils.getFieldVal(manager, "entries", false);
```
lukehutch: @eddi-weiss Question: This `manager.entries` field was working fine for earlier Equinox applications. Are the values returned by `manager.fragments` always a superset of `manager.entries`, and is that true with older versions of Equinox as well as recent versions? Shouldn't classpath entries be pulled in from both fields to be safe?

PR #323, opened 2019-02-25
321: Illegal method access behavior

eddi-weiss:


src/test/java/io/github/test/classgraph/issues/issue321/ClassInfoListTest.java
```diff
@@ -0,0 +1,35 @@
...
+import io.github.classgraph.ScanResult;
+
+public class ClassInfoListTest
+{
+
+   @Test
+   public void testNames()
+   {
+      try (ScanResult result = new ClassGraph()
+         .removeTemporaryFilesAfterScan()
```
lukehutch: `.removeTemporaryFilesAfterScan()` is not needed in general, whenever you call `ScanResult#close()`, temporary files are deleted. This is only needed if you think the `ScanResult` may not be appropriately closed.

PR #344, opened 2019-05-08
Fixed exception during loading of method with array-type arguments

ashcheglov:
Hi Luke,
Added class name to ArrayTypeSignature accordingly to https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getName() (actually, it is just parser input)
Tested on Java 11

Regards


lukehutch:
@ashcheglov Thank you so much for not just reporting the bug, but also providing the fix, and a corresponding unit test!

lukehutch:
@ashcheglov I modified your PR to use the human-readable form of the type name (`int[][]`) rather than the raw type name (`[[I`), and extended this by adding a new `ArrayClassInfo` that is a subclass of `ClassInfo` to represent array classes. This is released in 4.8.31. Thanks!

Issue #345, opened 2019-05-11
classfileResource not set for superclasses


cdietrich:
i use Classgraph for JDK classes.
i have the following problem
if a subclass is scanned first and the parent class after
the superclass entry will have `classfileResource == null`
as of `io.github.classgraph.ClassInfo.getOrCreateClassInfo(String, int, Map<String, ClassInfo>)`

my expectation would be that 
`io.github.classgraph.ClassInfo.addScannedClass(String, int, boolean, Map<String, ClassInfo>, ClasspathElement, Resource)`
corrects this in the else branch, (`classInfo.classfileResource = classfileResource;`)
but it does not

cdietrich:
```
public class Main {
	
	public static void main(String[] args) {
        ScanResult result = new ClassGraph()
        		.verbose()
                .enableClassInfo()
                .enableSystemJarsAndModules()
                .scan();
        for (ClassInfo i : result.getAllClasses()) {
        	if (i.getName().equals("java.util.AbstractQueue")) {
        		System.out.println(i.getName());
        		System.out.println(i.getResource());
        	}
        }
    }

}
```

cdietrich:
[classgraph.log](https://github.com/classgraph/classgraph/files/3169401/classgraph.log)

cdietrich:
the same may hold true for classmodifiers

lukehutch:
@cdietrich Released in 4.8.32. Thank you for the bug report and the pull request containing the fix and the testcase!

lukehutch:
PS I added the fix for the modifier bits, thanks for pointing that out too.

cdietrich:
@lukehutch  looks like inner classes are handled differently and have the problem too

```
	private static class Base {
		
	}
	
	public static class Sub extends Base {
		 
	}
	
	@Test
	public void issue345b() {
		try (ScanResult scanResult = new ClassGraph()
				.enableClassInfo()
				.overrideClassLoaders(
						new URLClassLoader(new URL[] { 
								Issue345.class.getResource("/io/github/classgraph/issues/issue345/Issue345$Sub.class"),
								Issue345.class.getResource("/io/github/classgraph/issues/issue345/Issue345$Base.class"),
								Issue345.class.getResource("/io/github/classgraph/issues/issue345/Issue345.class"),
						}))
				.scan()) {
			ClassInfo classInfo = scanResult.getClassInfo("io.github.classgraph.issues.issue345.Issue345$Base");
[...]

PR #346, opened 2019-05-11
Fixed classfileresource of classes that were created as superclass before they where scanned.

cdietrich:
Fixes https://github.com/classgraph/classgraph/issues/345


cdietrich:
i am not sure how to deal with modifiers. do they need to be set / corrected too?

lukehutch:
@cdietrich Thanks, adding a back-link from `ClassInfo` to classfile `Resource` was a fairly recent addition, and yes, the line you added is the correct fix. I'll take a look at the modifiers issue.

PR #349, opened 2019-05-20
Fix classpath detection for nested ZipBundleFile

larsgrefer:
If a `NestedBundleFile` contains a `ZipBundleFile` their paths must be concatenated using `!/` instead of `/`.


lukehutch:
Thank you for the pull request, @larsgrefer! LGTM. Released in 4.8.37.

larsgrefer:
Maybe the whole method should be rewritten in order to test for different `BundleFile` subclasses instead of blindly trying some field names.

lukehutch:
@larsgrefer It's a good suggestion -- want to take a shot at it? :-)

Work on `ClassLoaderHandler` implementations in particular I need external contributions for, since I am not familiar with most of the runtime environments out there (even how to launch code on a given runtime environment).

Issue #350, opened 2019-06-04
Include superclass fields in ScanResult


reitzmichnicht:
I'm a really happy user of this library, but stumpled upon a missing feature.

When calling .getClassesWithFieldAnnotation(String className) on the ScanResult it will only look at the declared fields of the class itself, not on any fields of the upper hierarchy.

I know this is the default java reflection behavior, but it would be either great to have an additional option to include upperclasses, or to change the default behavior because in my opinion a class should always be treated including all inherited stuff.

Kind regards,
Michael

lukehutch:
Hi @reitzmichnicht,

This is a good suggestion. I modified most methods to work with class inheritance many months ago, and added `.getDeclared...()` versions for when you want only matches in the class itself, but I missed this  `getClassesWithFieldAnnotation` method, and also  `getClassesWithMethodAnnotation`. 

It would actually be a simple fix: before these methods return, if the field or method is not private, and the class is not final, add into the result list all subclasses of the class. (Then also create a `.getDeclared...()` version that does not do this.)

Want to take a shot at it?

lukehutch:
Actually it's not quite as straightforward as I originally thought, so I went ahead and implemented a prototype. Can you please test the git master version and let me know if it works for you? Thanks.

reitzmichnicht:
Yes I already had a look and did not find an easy point for extension.
I tried your snapshot version, but I realized that it doesn't help for my case:

```
public class IdEntity {
  @Id
  @Column(updatable = false)
  @GeneratedValue
  private Long id;
}

public class FooEntity extends IdEntity {
  ...
}

  private static Stream<? extends Class<?>> findGeneratedEntities() {
    return new ClassGraph()
        .whitelistPackages(EntityTest.class.getPackage().getName())
        .enableFieldInfo()
        .enableAnnotationInfo()
        .ignoreFieldVisibility()
        .scan()
        .getClassesWithFieldAnnotation(GeneratedValue.class.getName())
        .filter(x -> !x.isAnonymousInnerClass() && !x.isAbstract() && x.isPublic())
        .loadClasses()
        .stream();
  }
```

If I look at your commit I expect that you do not consider private fields of superclasses which is normally correct. Is it reasonable to consider them when the ignoreFieldVisibility() option is set?

lukehutch:
[...]

PR #351, opened 2019-06-12
Change visibility of EMPTY_LIST to use it for initialization.

reitzmichnicht:
This change is to improve chainability of ClassInfoList results when combining multiple together. Does not affect immutability of ClassInfoList.

Old code:
```
    ClassInfoList classes = null;
    try (ScanResult scanResult = new ClassGraph()
        .whitelistPackages(testClass.getPackage().getName())
        .scan()) {
          for (Class<?> clazz : this.ofTypes) {
            if (clazz.isInterface()) {
              ClassInfoList cil = scanResult
                  .getClassesImplementing(clazz.getName());
              classes = classes == null ? cil : classes.union(cil);
            } else {
              ClassInfoList cil = scanResult
                  .getSubclasses(clazz.getName());
              cil.add(scanResult.getClassInfo(clazz.getName()));
              classes = classes == null ? cil : classes.union(cil);
            }
          }
    }
```

New code:
```
    ClassInfoList classes = ClassInfoList.EMPTY_LIST;
    try (ScanResult scanResult = new ClassGraph()
        .whitelistPackages(testClass.getPackage().getName())
        .scan()) {
          for (Class<?> clazz : this.ofTypes) {
            if (clazz.isInterface()) {
              ClassInfoList cil = scanResult
                  .getClassesImplementing(clazz.getName());
              classes = classes.union(cil);
            } else {
              ClassInfoList cil = scanResult
                  .getSubclasses(clazz.getName());
              cil.add(scanResult.getClassInfo(clazz.getName()));
              classes = classes.union(cil);
            }
          }
    }
```

### DIFF
--- a/src/main/java/io/github/classgraph/ClassInfoList.java
+++ b/src/main/java/io/github/classgraph/ClassInfoList.java
@@ -124,7 +124,7 @@ public class ClassInfoList extends MappableInfoList<ClassInfo> {
     }
 
     /** An unmodifiable empty {@link ClassInfoList}. */
-    static final ClassInfoList EMPTY_LIST = new ClassInfoList() {
+    public static final ClassInfoList EMPTY_LIST = new ClassInfoList() {
         @Override
         public boolean add(final ClassInfo e) {
             throw new IllegalArgumentException("List is immutable");


lukehutch:
@reitzmichnicht This is reasonable, but I think that a method `ClassInfoList::emptyList()` should be added instead, consistent with `Collections::emptyList()`. The `Collections::emptyList()` method is almost always to be preferred over `Collections::EMPTY_LIST`, because `Collections::emptyList()` does parameter type inference. In this case, `ClassInfoList` is not a generic class, but the habit of avoiding `EMPTY_LIST` is still probably a good one. I added the method.

Also, as I mentioned in #350, using `ClassInfoList::union(ClassInfoList)` iteratively in this way is a bad idea performance-wise, because the performance will scale as O(N^2). It's fine for a small project, but your performance can completely tank given a large result set.

Also, as also mentioned in #350, you shouldn't use `ClassInfo` objects outside the `try` block, since the `ScanResult` is closed (so `ClassInfo::loadClass()` won't work, and other things might break).

Instead, what you should do is:

```java
    try (ScanResult scanResult = new ClassGraph()
        .whitelistPackages(testClass.getPackage().getName())
        .scan()) {
          Set<ClassInfo> classes = new HashSet<>();
          for (Class<?> clazz : this.ofTypes) {
            if (clazz.isInterface()) {
              ClassInfoList cil = scanResult
                  .getClassesImplementing(clazz.getName());
              classes.addAll(cil);
            } else {
              ClassInfoList cil = scanResult
                  .getSubclasses(clazz.getName());
              classes.add(scanResult.getClassInfo(clazz.getName()));
              classes.addAll(cil);
            }
          }
          // Use `classes` here
    }
```
```java
```

lukehutch:
Released in 4.8.39, as an `emptyList()` method. I also exposed the constructors of `ClassInfoList`.

https://github.com/classgraph/classgraph/releases/tag/classgraph-4.8.39

However be aware that calling `ClassInfoList::add(ClassInfo)` will *not* deduplicate classes that are added to the list. So you should use the `Set<ClassInfo>::addAll` pattern I showed in the previous comment.

PR #359, opened 2019-07-24
maven-compiler-plugin 3.8.1

sullis:


lukehutch:
Thanks for this PR! There have been several bugs in 3.8.0 that I had to work around, and it took forever for them to release 3.8.1, so hopefully they fixed the issues I ran into.

PR #361, opened 2019-07-29
rename some tests not run by maven

isopov:
I have found that some tests in projects does not start or end with `Test` and so are not run by maven. So there are `Tests run: 127` in report, and after renaming tests `Tests run: 138`.

Unfortunately one of these tests is not stable - it relies on not standardized implementation-specific `toString()` implementation of Annotation from stdlib. Locally I have it passing in the IDE, but failing from terminal (seemingly on one and the same JDK, but maybe maven catches some other)

Is this situation an accident?

### DIFF
--- a/src/test/java/io/github/classgraph/features/AnnotationEqualityTest.java
+++ b/src/test/java/io/github/classgraph/features/AnnotationEqualityTest.java
@@ -16,7 +16,7 @@ import io.github.classgraph.ScanResult;
 /**
  * AnnotationEquality.
  */
-public class AnnotationEquality {
+public class AnnotationEqualityTest {
     /**
      * The Interface W.
      */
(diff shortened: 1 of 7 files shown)


isopov:
Here is the build for this PR on Travis (BTW, why don't you enable Travis check for PRs?)
https://travis-ci.org/classgraph/classgraph/builds/564975991
It fails on java11, but passes on java8 and has `Tests run: 138`
Latest build for master - https://travis-ci.org/classgraph/classgraph/jobs/562952373 has `Tests run: 127`

lukehutch:
Thanks. The tests are all run fine on Eclipse, so I didn't notice Maven fails to run all the tests.

I will replace all toString() tests -- I didn't know there was so much variability between implementations.

Travis does already run tests on PRs. I don't know why the tests weren't showing up here. Do you have any idea?

PR #362, opened 2019-07-29
migrate to junit5

isopov:


lukehutch:
@isopov Many thanks for doing this work to port ClassGraph to junit5!

PR #365, opened 2019-08-14
add support for last modified and posix file perms - fixes #364

### LINKED ISSUE #364: Feature Request: File Permissions in Resource
I'm extracting the contents of the JARs and need to preserve file permissions (read, write, exec).  The info is available via `ZipArchiveEntry::getUnixMode` but I'd need that info in `Resource`.

Thanks for this great library!

jamesward:


src/main/java/nonapi/io/github/classgraph/fastzipfilereader/LogicalZipFile.java
```diff
@@ -609,11 +613,51 @@ private void readCentralDirectory(final ZipFileSliceReader zipFileSliceReader, f
                 }
                 final boolean isDeflated = compressionMethod == /* deflated */ 8;
 
+                final int lastModifiedTime = entryBytes != null ? ZipFileSliceReader.getShort(entryBytes, entOff + 12)
+                        : zipFileSliceReader.getInt(cenPos + entOff + 12);
```
lukehutch: You've got `getShort` followed by `getInt` here -- this is probably incorrect, since the `entryBytes != null` test is just to test whether to use memory-mapped IO or not. There are several other cases of this below.

jamesward: Good catch.  Fixing now.

src/main/java/nonapi/io/github/classgraph/fastzipfilereader/LogicalZipFile.java
```diff
@@ -609,11 +613,51 @@ private void readCentralDirectory(final ZipFileSliceReader zipFileSliceReader, f
...
+
+                // MS-DOS Date & Time Format
+                final int lastModifiedSecond = (lastModifiedTime & 0b11111) * 2;
+                final int lastModifiedMinute = lastModifiedTime >> 5 & 0b111111;
+                final int lastModifiedHour = lastModifiedTime >> 11;
+                final int lastModifiedDay = lastModifiedDate & 0b11111;
+                final int lastModifiedMonth = (lastModifiedDate >> 5 & 0b111) - 1;
+                final int lastModifiedYear = (lastModifiedDate >> 9) + 1980;
+
+                final Calendar lastModifiedCalendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
```
lukehutch: Zipfiles don't have a timezone recorded anywhere in them -- so I guess it is it fair to pick UTC... Do you know what different zip tools do with this? Is UTC an ad hoc standard for zip entry timestamps?

src/main/java/io/github/classgraph/ClasspathElementZip.java
```diff
@@ -281,7 +281,7 @@ void open(final WorkQueue<ClasspathEntryWorkUnit> workQueue, final LogNode log)
      * @return the resource
      */
     private Resource newResource(final FastZipEntry zipEntry, final String pathRelativeToPackageRoot) {
-        return new Resource(this, zipEntry.uncompressedSize) {
+        return new Resource(this, zipEntry.uncompressedSize, zipEntry.lastModified, zipEntry.posixFilePermissions) {
```
lukehutch: `ClasspathElementDir` needs corresponding changes, and then I think the PR is ready to be merged. I can jump in and do that, or let me know if you're on a roll :-)

jamesward: Cool. I’ll get that done tomorrow.

src/main/java/nonapi/io/github/classgraph/fastzipfilereader/LogicalZipFile.java
jamesward: AFAICT the only viable option is UTC but details are hard to find.

src/main/java/io/github/classgraph/ClasspathElementZip.java
lukehutch: In that case, I'll go ahead and make the change in `ClasspathElementDir`, that way I can push out a release tonight, and you'll have a jar you can use in Maven Central by tomorrow.

src/main/java/nonapi/io/github/classgraph/fastzipfilereader/LogicalZipFile.java
lukehutch: UTC is the only reasonable default, so this is fine. I updated the JavaDoc (after pushing out the release though) to point out this possibly erroneous default assumption.

PR #383, opened 2019-11-23
Sort fields for deterministic order

contextshuffling:
Test in `JSONSerializationTest` depends on the fields return by `java.Class.getDeclaredFields`. Thus, the serialization result might change if it returns a different order of fields.

However, java.lang.Class.getDeclaredFields does not guarantee any specific order and thus, test can fail if the order is different, (i.e., it generates a different hash code). "The elements in the returned array are not sorted and are not in any particular order" (reference: https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredMethods--)

This PR proposes to fix it by:
* Sorting fields for a deterministic order.
* Make sure "format" is always the first key, because `ScanResult` class has a hard-coded regex that expects `format` as the first key.


lukehutch:
Thanks @contextshuffling for the PR! I modified it slightly to cache the `Comparator`, and to only put a field named `format` first if the class was `SerializationFormat` (which is the toplevel class that ClassGraph uses to serialize to JSON).

Just curious -- what platform and JRE version did the unit test fail on for you?

PR #389, opened 2019-11-28
Enable remote jar scanning for URLs inside URLClassLoader.

chrisr3:
This allows scanning of URLs that use a custom scheme.

I've refactored `CustomURLScheme` into a separate class because
`URL.setURLStreamHandlerFactory()` can only be invoked _once_ for a given JVM.


lukehutch:
@chrisr3 thanks for the patch!

Issue #400, opened 2020-02-08
OutOfMemoryError spins library out of control


TWiStErRob:
## Issue summary
`nonapi.io.github.classgraph.fastzipfilereader.NestedJarHandler#close` may never complete if Thread was interrupted, which is always if scanning failed. See below for details.

## Background
I was writing some unit tests for a really basic Neo4J graph, but while Neo4J-OGM was looking for my Entities during test setup, it just never completed. With some very roundabout debugging (all the async and OOM interplay was making me tear my hair out), I managed to narrow down the root cause to how classgraph works. I have 32G memory, and Java defaults to using 1/4th of memory per process, which means I cannot run much in parallel, so I have a global 256M override for each process. It works fine for Gradle, Neo4J, and other traditionally resource-hungry setups. It even works for my test setup most of the time, so for this report repro I lowered the memory limit to be consistent.

## Repro Setup
trivial library usage:
```java
	// implementation "io.github.classgraph:classgraph:4.8.62"
	ScanResult result = new ClassGraph()
			.whitelistPackages("pack.age")
			.verbose()
			.scan()
			;
```
plus a dependency with a nested jar in it:
```gradle
runtimeOnly "org.neo4j:neo4j-lucene-upgrade:3.4.9"
```
non-trivial: resource constrained memory: `-Xmx128M` (see build.gradle)

Full minimal repro can be found here: https://github.com/TWiStErRob/repros/tree/master/classgraph/oom-NestedJarHandler-spin

## Repro steps
1. `gradlew run`

or

1. Import project to IntellIJ
1. Execute `Main.main` (should pass, depending on machine setup)
1. Edit Run Configuration to have VM options: `-Xmx128M`
1. Run or debug again (will fail, expected: print a line)

[gradlew run.log](https://github.com/classgraph/classgraph/files/4174299/run.log)

## Pieces of the puzzle
What I found to be contributing to the issue.

### Nested Jar files
```gradle
runtimeOnly "org.neo4j:neo4j-lucene-upgrade:3.4.9"
```
an example from Neo4J/OGM, which contains !/lib/lucene-backward-codecs-5.5.0.jar
This is required to trigger
```java
                    } else {
                        // This path has one or more '!' sections.
```
code path in `NestedJarHandler`.

### Large allocation to trigger OOM
[...]

PR #401, opened 2020-02-09
Add performance regression test for #400

TWiStErRob:


src/test/perf/io/github/classgraph/issues/issue400/Issue400.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * @return used JVM heap size allocated in RAM
+     * @see <a href="https://stackoverflow.com/a/42567450/253468">What are Runtime.getRuntime().totalMemory() and freeMemory()?</a>
+     */
+    private long usedRam() {
+        Runtime runtime = Runtime.getRuntime();
+        runtime.gc();
+        runtime.gc();
+        runtime.gc();
```
lukehutch: It's worth calling `System.runFinalization();` after GC, since that makes it more sure that GC will complete and free available resources. I'll make that change though.

lukehutch:
Many thanks for submitting this PR!

PR #406, opened 2020-03-21
Added support for Quarkus 1.3

mcollovati:
Since version 1.3 Quarkus changed how classloading works.
This change adds support for the new Quarkus classloader
and removes unecessary entries added for 1.2 Runtimeclassloader.

### DIFF
--- a/src/main/java/nonapi/io/github/classgraph/classloaderhandler/QuarkusClassLoaderHandler.java
+++ b/src/main/java/nonapi/io/github/classgraph/classloaderhandler/QuarkusClassLoaderHandler.java
@@ -30,7 +30,6 @@ package nonapi.io.github.classgraph.classloaderhandler;
 
 import java.nio.file.Path;
 import java.util.Collection;
-import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


lukehutch:
Looks good to me. I really appreciate the contribution!

I'm curious how ClassGraph actually operates within Quarkus, since Quarkus runs on GraalVM and involves all sorts of magic like partial evaluation. Does the runtime environment still look like a normal JVM runtime environment from the point of view of a program compiled for Quarkus?

Please report any issues you have with ClassGraph in Quarkus. GraalVM/Quarkus are exciting projects.

mcollovati:
I'm not a Quarkus expert. It seems to me that GraalVM comes into play only when building native images.
I managed to make ClassGraph work on development mode and as standard and uber jar; however it doesn't work for native image.
I think a Quarkus extension should be implemented to put in the magic you mentioned above; I'm trying to learn more about this stuff.

lukehutch:
Makes sense. If building a native app, you would probably need to use build-time scanning (the ClassGraph wiki has a page on this).

mcollovati:
I missed that page; thanks for pointing out

Issue #408, opened 2020-03-27
OSGi Optional Packages


skahmann:
The following packages are set as required imports in the pom:

- jdk.internal.misc
- sun.misc
- sun.nio.ch

These should be optional to properly support OSGi runtimes:

`;resolution:="optional"`

lukehutch:
See discussion in #409.

lukehutch:
Released in 4.8.66. Thanks again!

PR #409, opened 2020-03-27
408: Added optional resolution to jdk and sun packages

skahmann:
Closes [408](https://github.com/classgraph/classgraph/issues/408)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -504,7 +504,7 @@
                                 <Require-Capability>osgi.ee;filter:="(&amp;(osgi.ee=JavaSE)(version=1.7))"</Require-Capability>
                                 <Export-Package>io.github.classgraph;version="${project.version}"</Export-Package>
                                 <!-- N.B. make sure the module-info "requires" entries match these imports -->
-                                <Import-Package>javax.xml.xpath,javax.xml.namespace,javax.xml.parsers,org.w3c.dom,jdk.internal.misc,sun.misc,sun.nio.ch</Import-Package>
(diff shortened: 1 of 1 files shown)


lukehutch:
@skahmann Thanks for this contribution. Will I need to replace all uses of classes in these packages with reflective access and a sane fallback, for OSGi environments that do not contain these packages?

lukehutch:
@skahmann Actually I checked, and all the access to these classes is already via reflection, with a sane fallback (requiring garbage collection to unmap byte buffers, rather than explicitly unmapping them).

However, the module descriptor still uses:

```
    // VersionFinder requires java.xml
    requires java.xml;
    // FileUtils requires jdk.unsupported (for usage of Unsafe)
    requires jdk.unsupported;
    // ModulePathInfo requires java.management
    requires java.management;
    // LogNode requires java.logging
    requires java.logging;
```

Replacing these with `requires static` may make ClassGraph projects fail even if the module is available, if some other module in the classpath has not already `requires`-ed the module.

Is it safe to assume that OSGi projects will never use JPMS modules, and continue using the legacy classpath?

skahmann:
I did some testing using Apache Karaf and everything seems to be working okay. I think it's safe to assume the OSGi runtime makes these classes available. I was tuned into this fix based on this issue: https://github.com/joel-costigliola/assertj-core/issues/1663

lukehutch:
@skahmann Thanks. Ș do this question:

> Is it safe to assume that OSGi projects will never use JPMS modules, and continue using the legacy classpath?

Issue #414, opened 2020-04-09
ClassGraph on Open Liberty doesn't find any classes from the application.


michael-simons:
Here's Michael from Neo4j-OGM project.

A customer reached out to us wanting to run OGM on Websphere Liberty respectively Open Liberty https://openliberty.io.

Turns out, that Neo4j-OGM doesn't find any domain classes in that scenario. After a day of pulling my hair, I'm relatively sure that I can boil this down to interactions of ClassGraph and Liberty.

I followed this guide https://openliberty.io/guides/maven-intro.html and added ClassGraph.

Given the following servlet

```
package io.openliberty.guides.hello;

import io.github.classgraph.ClassGraph;
import io.github.classgraph.ScanResult;

import java.io.IOException;

import javax.servlet.ServletException;
import javax.servlet.annotation.WebServlet;
import javax.servlet.http.HttpServlet;
import javax.servlet.http.HttpServletRequest;
import javax.servlet.http.HttpServletResponse;

@WebServlet(urlPatterns="/servlet")
public class HelloServlet extends HttpServlet {
    private static final long serialVersionUID = 1L;

    protected void doGet(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {

        int numberOfClassesFound = 0;
        // The servlet itself should be found... But isnt.
        try (ScanResult r = new ClassGraph().verbose(true).enableRealtimeLogging().enableAllInfo().whitelistPackages("io.openliberty.guides.hello").scan()) {
            numberOfClassesFound = r.getAllClasses().size();
        }


        response.getWriter().append("Number of classes " + numberOfClassesFound + "\n");
    }

    protected void doPost(HttpServletRequest request, HttpServletResponse response)
        throws ServletException, IOException {
        doGet(request, response);
    }
}
```

Should find at least 1 class, the servlet itself.
However, it does not.

Follow this step to recreate from the attached producer

```
unzip classgraph-on-liberty.zip -d reproducer
cd reproducer
mvn liberty:dev   
```

Hit any key after the thing starts or go to http://localhost:9080/ServletSample/servlet

You'll see 0 classes found respectively a failing test.

I have attached the ClassGraph verbose log as well. What sticks out is:

```

[...]

PR #415, opened 2020-04-11
Fix for Libery running in "overlay"

cpierceworld:
Fix for LIberty issue #414

In this fix, I attempt to use the methods that just return the classpath urls.   I can't regression test against any old Liberties so I have it fallback to the old logic whenever the new logic fails.

- Use SmartClassPath's "getClassPath" to get the classpath URL(s).
- If that doesn't work for some reason, revert to processing individual SmartClassPath elements.
  - Attempt to use "getContainerURLs" on SmartClassPath's elements
  - If that doesn't work, revert to introspection on the "container"
    - Attempt to use "getURls" on the container.
    - If that doesn't work,  revert to introspection on the "delegate".


lukehutch:
Awesome, thanks for this work, @cpierceworld! I really appreciate your work on it.

PR #422, opened 2020-04-30
Don't read through and store classpath elements if they won't be used…

### LINKED ISSUE #421: ClasspathFinder is reading bytes into memory and immediately throwing them away
I ran into some OOM problems with a very large parent classpath even when I set the parent to be ignored. I discovered that the following code in ClasspathFinder.java is not quite ignoring the parent.

```
                    final LogNode classloaderHandlerLog = classloaderURLLog == null ? null
                            : classloaderURLLog
                                    .log("Ignoring parent classloader " + classLoader + ", normally handled by "
                                            + classLoaderHandlerRegistryEntry.classLoaderHandlerClass.getName());
                    classLoaderHandlerRegistryEntry.findClasspathOrder(classLoader, ignoredClasspathOrder, scanSpec,
                            classloaderHandlerLog);
```
This reads everything in the parent for the where the user is  ignoring parents, but scanning the entire java.class.path anyway (?). There are a number of settings under which the retained information is never used, and this just becomes a memory usage liability. (PR shortly)

As commentary, I think I would find it confusing to sort out the case where a class on the class path was loaded by the parent and then mysteriously didn't show up when I scanned the java.class.path... so I find that use case a bit dubious to begin with, but I imagine someone must have it.

nsoft:
…. Fixes #421

### DIFF
--- a/src/main/java/nonapi/io/github/classgraph/classpath/ClasspathFinder.java
+++ b/src/main/java/nonapi/io/github/classgraph/classpath/ClasspathFinder.java
@@ -243,8 +243,10 @@ public class ClasspathFinder {
                             : classloaderURLLog
                                     .log("Ignoring parent classloader " + classLoader + ", normally handled by "
                                             + classLoaderHandlerRegistryEntry.classLoaderHandlerClass.getName());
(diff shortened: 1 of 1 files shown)


lukehutch:
Thanks for the PR. I might tweak the logic a bit (e.g. explicitly disable `java.class.path` if parent classloaders are disabled), but otherwise it looks good to me.

PR #424, opened 2020-05-01
initial UnoJarClassLoaderHandler implementation

nsoft:
I saw you already added something... PR for comparison. Specifically one thing to note vs your earlier commit is that the main/ dir is a location where main.jar is found (not class files)


lukehutch:
Thanks, that's pretty much identical to what I have, other than the `main/` part, so I'll close this PR and fix the `main/` issue. So there are no bare classfiles in the jar, with the package root rooted anywhere other than the root of the jar? And `main/` does not contain any classfiles outside `main.jar`?

Issue #436, opened 2020-06-13
Allow adding on additional URL Protocol Handlers


GedMarc:
**https://github.com/classgraph/classgraph/blame/a0d510593e3da9f41852a1210112a0f96aa3647e/src/main/java/io/github/classgraph/Resource.java#L104**

Please move this to the URL class to throw the exception - It is possible to apply a custom service loader for URL to handle JRT classes.

GedMarc:
I can't make a PR for it for the JDK 7 compatibility - requires java.nio.file.Path to create a URL Connection, 
May have to leave this to the caller?

lukehutch:
The whole point of the special error message for `jrt:` URLs was that I didn't want to throw `MalformedURLException` from `Resource.getURL()` (since this exception should almost never be thrown, with `jrt:` URLs being one exception that I found out about later).

However, it's possible to check if a URL scheme is registered, so I'll add a check for that.

lukehutch:
OK, so I made the simplest change possible, moving the `jrt:` scheme check inside the `catch (MalformedURLConnection e)` block.

This isn't really a big enough change to push out a new release. Really, if you need `jrt:`, you should be using `URI`, not `URL`.

PR #437, opened 2020-06-14
Remove JRT error and fallback to toURL()

GedMarc:
https://github.com/classgraph/classgraph/issues/436

Allow the consumer to create a URL Protocol for handling JRT URL's, Throw the URL exception if it doesn't find a valid protocol handler


lukehutch:
Just saw this -- obsoleted by most recent checkin.

PR #453, opened 2020-06-25
Add cause to IllegalArgumentException thrown by ScanResult.loadClass()

ujibang:
adding the cause to the IllegalArgumentException thrown by ScanResult.loadClass() allows to programmatically handle different error cases


lukehutch:
Thanks!

PR #491, opened 2020-12-23
Fix NPE in TypeArgument's equals and hashCode

Tagakov:
Since `typeSignature` in `TypeArgument` could be null it should be honored in equals and hashcode methods


lukehutch:
Thanks for the pull request, @Tagakov! Fixed and released as 4.8.98.

PR #492, opened 2020-12-24
ci: add JDK 15 to build matrix

sullis:


lukehutch:
Thanks @sullis! JDK 15 is the most important release in many years, it's way more efficient than previous releases.

Issue #543, opened 2021-08-10
ClassInfoList.getClassesImplementing() throws exception if underlying object is a class despite documentation stating the contrary


parttimenerd:
The documentation and the code differ for [`ClassInfoList.getClassesImplementing()`](https://github.com/classgraph/classgraph/blob/046a86d5f74c140e9d010d39a44b02fe3b94c17c/src/main/java/io/github/classgraph/ClassInfo.java#L1687):
```java
    /**
     * Get the classes (and their subclasses) that implement this interface, if this is an interface.
     *
     * @return the list of the classes (and their subclasses) that implement this interface, if this is an
     *         interface, otherwise returns the empty list.
     */
    public ClassInfoList getClassesImplementing() {
        if (!isInterface()) {
            throw new IllegalArgumentException("Class is not an interface: " + getName());
        }
```
The code clearly throws an `IllegalArgumentException` for non-interfaces.

This propagates to [`ScanResult.getClassesImplementing()`](https://github.com/classgraph/classgraph/blob/046a86d5f74c140e9d010d39a44b02fe3b94c17c/src/main/java/io/github/classgraph/ScanResult.java#L1031) too which therefore has the same issue.

lukehutch:
Thanks, good catch. Do you think this should throw an exception, or just return the empty list? The argument for throwing an exception is that it may help find bugs in a user's code, but maybe it is better never to throw exceptions, and fail gracefully instead.

parttimenerd:
I would personally prefer the latter. My current use case (where this came up) is getting all implementing classes and subclasses for a ClassInfo object (not caring about the distinction between interfaces and classes). 

A preliminary search on GitHub for usages of the method shows that most usages are in the context where the `ClassInfo` object is expected to be an interface.

lukehutch:
Fixed and released in 4.8.113. Thanks!

PR #549, opened 2021-08-15
Add typed overloads for methods taking class names

larsgrefer:
This allows a more typesafe way of working with scan results.

### DIFF
--- a/src/main/java/io/github/classgraph/AnnotationInfoList.java
+++ b/src/main/java/io/github/classgraph/AnnotationInfoList.java
@@ -28,6 +28,7 @@
  */
 package io.github.classgraph;
 
+import java.lang.annotation.Annotation;
 import java.lang.annotation.Repeatable;
 import java.util.ArrayList;
 import java.util.HashSet;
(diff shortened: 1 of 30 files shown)


lukehutch:
@larsgrefer thank you for this contribution!

A couple of things --

(1) You added some `Assert` checks to these methods, to ensure the parameter types were reasonable. There was a request recently to drop these sorts of checks. What do you think about this?: https://github.com/classgraph/classgraph/issues/543

(2) Actually using `Class<?>` references was standard practice in the early days of ClassGraph (back when it was called FastClasspathScanner), and I changed the entire API to use class names rather than class references, because using a class reference triggers classloading and class initialization, and I had several users run into problems where e.g. a superclass' `Class<?>` reference was loaded in one classloader, and all the subclasses found by FastClasspathScanner were loaded by a different classloader. This would mean that the subclass couldn't be cast to its own superclass without throwing a `ClassCastException`. ClassGraph has evolved a lot since then though, and it goes out of its way to *not* load or initialize classes. Additionally it tries to load classes with its own resource-based classloader only as a very last resort, if no other classloader can load a given class. So I think the situation is better now, and maybe these methods can be added back to ClassGraph.

larsgrefer:
(1) I'd argue that the Assert statements are just proper input validation, but we could add some javadoc to the methods explaining that an illegal argument will trigger an IllegalArgumentException.

(2) I don't want to force any classloading nor change the way how classgraph works internally. This PR only offers a way to better deal with cases, where some known classes are already loaded anyway.

lukehutch:
Yes, I think this is all very reasonable. Thanks again for the great code contribution. I'll merge it and push out a release shortly.

lukehutch:
Released in 4.8.115.

PR #550, opened 2021-08-16
Cleanup Code

larsgrefer:
My IDE found some possible code cleanups

### DIFF
--- a/src/main/java/io/github/classgraph/ObjectTypedValueWrapper.java
+++ b/src/main/java/io/github/classgraph/ObjectTypedValueWrapper.java
@@ -629,21 +629,21 @@ class ObjectTypedValueWrapper extends ScanResultObject {
         } else if (stringValue != null) {
             buf.append(stringValue);
         } else if (integerValue != null) {
-            buf.append(Integer.toString(integerValue));
+            buf.append(integerValue);
         } else if (longValue != null) {
-            buf.append(Long.toString(longValue));
+            buf.append(longValue);
         } else if (shortValue != null) {
-            buf.append(Short.toString(shortValue));
+            buf.append(shortValue);
         } else if (booleanValue != null) {
-            buf.append(Boolean.toString(booleanValue));
+            buf.append(booleanValue);
         } else if (characterValue != null) {
-            buf.append(Character.toString(characterValue));
+            buf.append(characterValue);
         } else if (floatValue != null) {
-            buf.append(Float.toString(floatValue));
+            buf.append(floatValue);
         } else if (doubleValue != null) {
-            buf.append(Double.toString(doubleValue));
+            buf.append(doubleValue);
         } else if (byteValue != null) {
-            buf.append(Byte.toString(byteValue));
+            buf.append(byteValue);
         } else if (stringArrayValue != null) {
             buf.append(Arrays.toString(stringArrayValue));
         } else if (intArrayValue != null) {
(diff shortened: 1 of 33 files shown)


src/main/java/io/github/classgraph/ObjectTypedValueWrapper.java
```diff
@@ -629,21 +629,21 @@ protected void toString(final boolean useSimpleNames, final StringBuilder buf) {
         } else if (stringValue != null) {
             buf.append(stringValue);
         } else if (integerValue != null) {
-            buf.append(Integer.toString(integerValue));
+            buf.append(integerValue);
```
lukehutch, line 632: Is there any chance of unboxing (leading to a call to `buf.append(int)` beating out the matching of `buf.append(Object)` here?

src/main/java/nonapi/io/github/classgraph/json/JSONUtils.java
```diff
@@ -76,7 +76,7 @@
             final char hexDigit1 = nibble1 <= 9 ? (char) ('0' + nibble1) : (char) ('A' + nibble1 - 10);
             final int nibble0 = c & 0xf;
             final char hexDigit0 = nibble0 <= 9 ? (char) ('0' + nibble0) : (char) ('A' + nibble0 - 10);
-            JSON_CHAR_REPLACEMENTS[c] = "\\u00" + Character.toString(hexDigit1) + Character.toString(hexDigit0);
+            JSON_CHAR_REPLACEMENTS[c] = "\\u00" + hexDigit1 + hexDigit0;
```
lukehutch, line 79: This looks dangerous to me... probably `+` will be applied left-to-right, but adding two `char`s that are supposed to be string-concatenated is dodgy, because integer addition will be applied if you're not careful

lukehutch:
Thanks for these changes! What IDE do you use? IntelliJ IDEA? Eclipse didn't catch these.

These changes all look reasonable, with the exception of a couple that I commented on that are probably fine but look a bit dodgy. I'll merge this then check on those ones I commented on.

src/main/java/io/github/classgraph/ObjectTypedValueWrapper.java
lukehutch: OK, Eclipse informs me that this does indeed apply `buf.append(Object)`. I guess I need to improve my knowledge of the semantics of boxing/unboxing and multiple dispatch!

src/main/java/nonapi/io/github/classgraph/json/JSONUtils.java
lukehutch: I rewrote this as `"\\u00" + hexDigit1 + "" + hexDigit0` just to be a bit safer / more explicit. This will probably be caught by a static analyzer too though as a no-op string concatenation.

lukehutch:
OK, these changes are all integrated. They don't change anything functionally, so I'll wait to push out another release. Thank you!

src/main/java/io/github/classgraph/ObjectTypedValueWrapper.java
larsgrefer: `integerValue` is of type `java.lang.Integer` (boxed) so `append(Object)` will be called.

lukehutch: Right, I wasn't sure in what order unboxing was applied relative to the selection of method to dispatch.

larsgrefer:
These were some of the default inspections of IntelliJ IDEA.

PR #592, opened 2021-11-06
`ClasspathFinder` supports overrides and `enableSystemJarsAndModules()` together

roxspring:
- Previously `enableSystemJarsAndModules()` was ignored if overriding the classpath / classloader. This change allows the two to be used together if desired. 
- Unamed modules from the parent classloader are excluded if the claspath / classloader is overridden. 
  - IIUC this was matches the comment and presumably the intended behaviour:
    > Only scan java.class.path if parent classloaders are not ignored, classloaders are not overridden, and the classpath is not overridden, unless only module scanning was enabled, and an unnamed module layer was encountered

### DIFF
--- a/src/main/java/nonapi/io/github/classgraph/classpath/ClasspathFinder.java
+++ b/src/main/java/nonapi/io/github/classgraph/classpath/ClasspathFinder.java
@@ -284,16 +284,15 @@ public class ClasspathFinder {
             classLoaderOrderRespectingParentDelegation = finalClassLoaderOrder.toArray(new ClassLoader[0]);
         }
 
+        // Never scan java.class.path if classloaders or classpath are overridden
+        if (scanSpec.overrideClassLoaders != null || scanSpec.overrideClasspath != null) {
+            return;
+        }
 
-
-        // Only scan java.class.path if parent classloaders are not ignored, classloaders are not overridden,
-        // and the classpath is not overridden, unless only module scanning was enabled, and an unnamed module
+        // Scan java.class.path if parent classloaders are not ignored, or if module scanning was enabled and an unnamed module
         // layer was encountered -- in this case, have to forcibly scan java.class.path, since the ModuleLayer
         // API doesn't allow for the opening of unnamed modules.
-        if ((!scanSpec.ignoreParentClassLoaders && scanSpec.overrideClassLoaders == null
-                && scanSpec.overrideClasspath == null)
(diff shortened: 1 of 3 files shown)


src/main/java/nonapi/io/github/classgraph/classpath/ClasspathFinder.java
```diff
@@ -121,7 +121,10 @@ public ClasspathFinder(final ScanSpec scanSpec, final LogNode log) {
         // If so, need to enable module scanning. If not, disable module scanning, since only the provided
         // classloader(s) should be scanned. (#382)
         boolean scanModules;
-        if (scanSpec.overrideClasspath != null) {
+        if (scanSpec.enableSystemJarsAndModules) {
+            // Always scan modules if system modules are enabled
+            scanModules = true;
+        } else if (scanSpec.overrideClasspath != null) {
```
lukehutch: Won't this mean that if `scanSpec.enableSystemJarsAndModules` is true, all modules will be scanned, not just system modules? I'm rusty on my recollection of how this all works, but I think there would need to be some additional logic here that checks if any modules have been accepted for scanning, and if not, scanning should be limited to just system modules.

roxspring: I think that's accurate: the `ModuleFinder` was indiscriminately finding all modules, whether system modules or not. 718bcba addresses this such that `ModuleFinder` only records system modules if `scanSpec.enableSystemJarsAndModules==true` and only records non-system modules if a new `scanNonSystemModules` parameter is true.

By the same logic, it only sets `forceScanJavaClassPath` if `scanNonSystemModules` parameter is true, based on my limited understanding that all system modules are named. Consequently the commit also reverts the changes around `java.class.path` scanning that are no longer necessary.

To be fair, I don't have an environment with non-system or unnamed modules so I haven't been able to test this logic outside of the test suite.

lukehutch:
@roxspring I really appreciate you digging into the complex logic here and submitting these changes, to try to find a good solution!

It looks good to me (and in a few places you were able to dramatically simplify the logic, which is great!). I'll merge the PR, mostly because I want to look at it all in one place in my IDE. I need to make sure it's not going to break anything for anyone. I think the logic is fine though.

lukehutch:
@roxspring I made a few minor tweaks, but overall the logic seems sound, and the tests all pass. I really appreciate your work on this!

Released in 4.8.130.

PR #594, opened 2021-11-08
fix AccessControlException from SecurityManager in ClasspathOrder

elkman:
Hi @lukehutch,

yes, it's me and SecurityManger again =)

We have a new SecurityException after upgrading from 4.8.69 to 4.8.129 with our SecurityManager enabled Tomcat target platform.

In ClasspathOrder, class path elements are converted to URLs by `pathElementURL = new File(pathElementToStr).toURI().toURL()`.

File#toUri() requires read access to the directory containing the file. If the JVM is running with SecurityManager enabled, this file system access may represent a security policy violation (e.g., in Tomcat, the `CATALINA_HOME/bin` folder containing classpath elements such as bootstrap.jar, etc.).
In this case, the existing fallback handling can be used.

This PR works fine in our case, but we are only searching for some WebJars.

### DIFF
--- a/src/main/java/nonapi/io/github/classgraph/classpath/ClasspathOrder.java
+++ b/src/main/java/nonapi/io/github/classgraph/classpath/ClasspathOrder.java
@@ -319,14 +319,18 @@ public class ClasspathOrder {
                 final String pathElementToStr = pathElement.toString();
                 try {
                     pathElementURL = new File(pathElementToStr).toURI().toURL();
-                } catch (final MalformedURLException e) {
+                } catch (final MalformedURLException | SecurityException e) {
+                    if (log != null) {
+                        log.log("Failed to convert classpath element to URL, "
+                                + "Try prepending \"file:\" to create a URL (" + e + "): " + pathElementStr );
+                    }
                     // Final fallback -- try prepending "file:" to create a URL
                     pathElementURL = new URL("file:" + pathElementToStr);
                 }
             }
-        } catch (final MalformedURLException e1) {
+        } catch (final MalformedURLException |SecurityException e1) {
             if (log != null) {
-                log.log("Cannot convert to URL: " + pathElement);
+                log.log("Cannot convert to URL (" + e1 + "): " + pathElement );
             }
             pathElementURL = null;
         }


lukehutch:
@elkman I can't believe people are still using a `SecurityManager` these days! You know it has been [deprecated for removal in JDK 17](https://openjdk.java.net/jeps/411), right? :-)

One of the most annoying thing about `SecurityManager` is the number of Java library calls that can throw `SecurityException` -- and the fact it's an unchecked exception. Do you know of any way to find all places in a project where `SecurityException` can be thrown by library code called by the project? There are probably more of these fixes that are needed...

lukehutch:
I found a few more places where `SecurityException` or `IllegalArgumentException` or `IOError` should be caught. Released in 4.8.131. Thank you for the pull request!

elkman:
Hi @lukehutch 
 
> @elkman I can't believe people are still using a `SecurityManager` these days! 

yes, it is quite annoying, but this particular datacenter (public sector) decided to use it for its managed Tomcat hosting, which includes an option to get a Tomcat instance on a shared VM (like PHP in the 90s, the times without VMs, containers, SELinux and AppArmor). 

> You know it has been [deprecated for removal in JDK 17](https://openjdk.java.net/jeps/411), right? :-)

In fact, `AccessControlException` is already deprecated with JDK 17, `SecurityException` will remain. So at least there is no need to catch deprecated exceptions.

> One of the most annoying thing about `SecurityManager` is the number of Java library calls that can throw `SecurityException` -- and the fact it's an unchecked exception. Do you know of any way to find all places in a project where `SecurityException` can be thrown by library code called by the project? There are probably more of these fixes that are needed...

No, I don't know and I also think that there is no easy way to find all missed `RuntimeExeptions`, because if programmers should care too much about them, they were not derived from `RuntimeExeption`.

I could think of a test that sets up a `SecurityManager` for some common cases of missing read permissions for the current directory, a jar file, a directory of a jar file and maybe some more. But I am not sure if it is worth the time spent on an already almost dead feature like the `SecurityManager`.

Thank you for the quick response and all your work on classgraph!

lukehutch:
You're welcome, and all good points. OK, I think I caught most of them. I searched for all uses of `Path.toUri`, `new URL`, `new URI`, `URI.toURL`, etc.

We can just leave it at that for now, but please report any further problems. Thanks!

PR #597, opened 2021-11-12
Mark resolution of jdk.internal.misc as optional (again)

tobias--:
jdk.internal.misc was marked optional in https://github.com/classgraph/classgraph/issues/408. I think this resolution attribute was unintentionally removed, when jvm.driver was added with this commit: https://github.com/classgraph/classgraph/commit/4c30481ada042af58f7ad03292fc411e276dfc47

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -489,7 +489,7 @@
                                 <Export-Package>io.github.classgraph;version="${project.version}"</Export-Package>
                                 <!-- Make sure "requires" entries in the module descriptor match these imports: -->
                                 <!-- src/main/module-info/io.github.classgraph/module-info.java -->
-                                <Import-Package>javax.xml.xpath,javax.xml.namespace,javax.xml.parsers,org.w3c.dom,jdk.internal.misc,io.github.toolfactory.jvm;resolution:="optional",sun.misc;resolution:="optional",sun.nio.ch;resolution:="optional"</Import-Package>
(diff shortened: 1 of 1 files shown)


lukehutch:
Thanks, sorry about that...

Out of curiosity, what broke due to this change?

lukehutch:
Fixed and released in 4.8.132

tobias--:
I tried using this bundle in an osgi environment for the first time. Resolving using bndtools failed, since bnd did not find any bundle that exports jdk.internal.misc. So I checked whether the package jdk.internal.misc has always been required. 

Is this package actually required by classgraph? I did not find any usage of this package. The "internal" also suggests that it is not supposed to be used by third parties.

lukehutch:
The `sun.misc.Unsafe` class is used by ClassGraph. `Unsafe` is used to unmap mapped ByteBuffers as soon as they are no longer needed, so that file handles are closed as early as possible on Windows (otherwise temporary files can't be deleted). There is a class `jdk.internal.misc.Unsafe` that just proxies calls to the Sun class. ClassGraph used to try to use the jdk version of the class if it was available, but if I remember right, the jdk version was missing some critical functionality that was in the sun version, so I dropped support for the jdk version, and just use the sun version. I must have forgotten to remove the jdk package from the Manifest.

https://github.com/classgraph/classgraph/blob/latest/src/main/java/nonapi/io/github/classgraph/utils/FileUtils.java#L499

The sun version of `Unsafe` looks like it will stay in place until both versions of this class are completely removed from the JDK, probably around JDK 21 or so would be my guess -- so using the sun version should be sufficient.

I guess I should remove this entry from the manifest entirely!

lukehutch:
Sorry, there was a regression in 4.8.132, please try 4.8.133 instead.

PR #600, opened 2021-11-19
Closing ClassfileReader must also close its underlying Resource.

chrisr3:
Upgrading from 4.8.90 to 4.8.133, I have noticed that the resources returned by `scanResult.allResources` are now in an "open" state, such that invoking `open()` on them throws an `IOException` with the message:
> "Resource is already open -- cannot open it again without first calling close()"

I believe I have a fix for this, and am currently trying to write some test cases.

### DIFF
--- a/src/main/java/io/github/classgraph/ClasspathElementFileDir.java
+++ b/src/main/java/io/github/classgraph/ClasspathElementFileDir.java
@@ -155,7 +155,6 @@ class ClasspathElementFileDir extends ClasspathElement {
                         "Skipping classpath element, since dir cannot be accessed: " + classpathEltDir, log);
             }
             skipClasspathElement = true;
-            return;
         }
     }
 
@@ -177,7 +176,7 @@ class ClasspathElementFileDir extends ClasspathElement {
             private FileSlice fileSlice;
 
             /** True if the resource is open. */
-            protected AtomicBoolean isOpen = new AtomicBoolean();
+            private final AtomicBoolean isOpen = new AtomicBoolean();
 
             @Override
             public String getPath() {
@@ -247,7 +246,7 @@ class ClasspathElementFileDir extends ClasspathElement {
                 // Classfile won't be compressed, so wrap it in a new FileSlice and then open it
                 fileSlice = new FileSlice(resourceFile, nestedJarHandler, /* log = */ null);
                 length = fileSlice.sliceLength;
-                return new ClassfileReader(fileSlice);
+                return new ClassfileReader(fileSlice, onClose());
             }
 
             @Override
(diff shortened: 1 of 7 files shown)


lukehutch:
Hi @chrisr3 -- thanks for this pull request and the unit test! 

These resources are in fact supposed to be closed after a classfile has been read, so I'm wondering if there's another fix that needs to be applied to fix the underlying failure to close the resource. I'll use your unit test to try to see what's actually failing here.

lukehutch:
For reference, here is where `ClassFileReader` instances are closed (by try-with-resources):

https://github.com/classgraph/classgraph/blob/latest/src/main/java/io/github/classgraph/Classfile.java#L1924

So I think the problem is that `ClassFileReader#close()` does not close the underlying `Slice`.

lukehutch:
Actually I was wrong in my assessment -- your fix is the correct fix. Thank you!

lukehutch:
@chrisr3 This is the correct fix -- thank you for this pull request, and for the unit test! Released in version 4.8.134.

chrisr3:
> @chrisr3 This is the correct fix -- thank you for this pull request, and for the unit test! Released in version 4.8.134.

Thanks @lukehutch, although I wasn't _quite_ expecting it to be merged just yet. I think there is still an issue with [ClasspathElementModule.java](src/main/java/io/github/classgraph/ClasspathElementModule.java) because closing the `InputStream` returned by its `Resource.open()` method will not yet reset its `isOpen` field.

lukehutch:
@chrisr3 Good catch, please see the latest commit above, and let me know if this resolves your concern.

I'll check the rest of the `Resource` implementations to make sure they're not missing any more resource close actions.

chrisr3:
Hi @lukehutch,

Not really, because the `super.close()` in `Resource.close()` should handle closing the `InputStream` already. My point was more that `InputStream.close()` doesn't close the `Resource` as it would for the other `Resource` types.

Chris

lukehutch:
@chrisr3 Yes, I just realized this and removed the last commit :-)

I understand the concern now. I have an InputStream wrapper in ClassGraph that allows a custom close action to be added. I'll wrap it in that.

lukehutch:
@chrisr3 OK, closing the `InputStream` should close the underlying `Resource` now.

However I noticed another problem: the `ByteBuffer` returned by `ModuleReaderProxy#read(String path)` does not close the underlying `Resource` either when `ModuleReaderProxy#release(ByteBuffer)` is called.

Last time I tried implementing a custom `ByteBuffer` (which is a ton of work!), I got all the way to the end of the process and then discovered that a critical method is non-public (I forget which one) -- so it is not possible to extend `ByteBuffer` in user code, it is only possible for the JDK team to add new `ByteBuffer` instances. I may need to change the ClassGraph API to return a `ByteBufferWrapper` instance that holds a `ByteBuffer` and also a reference to the underlying `Resource`, so that that can be closed.

lukehutch:
I extended the `Resource` interface to include a `readCloseable()` method that returns a `CloseableByteBuffer` that calls `Resource#close()` when `CloseableByteBuffer#close()` is called. I think this is the best that can be done, since `ByteBuffer` cannot be extended (the problem is that `ByteBuffer` has no protected constructors).

Released these fixes in version 4.8.135. Thanks!

PR #602, opened 2021-11-22
Only invoke Resource's super.close() if the resource is open.

chrisr3:
Protect `Resource`'s `super.close()` invocation using `Resource.isOpen`. This allows us to simplify the `onClose` handlers only to invoke `Resource.close()`.

Maybe one day the `onClose` handler can then be replaced by just `Resource::close`.

### DIFF
--- a/src/main/java/io/github/classgraph/ClasspathElementFileDir.java
+++ b/src/main/java/io/github/classgraph/ClasspathElementFileDir.java
@@ -182,9 +182,7 @@ class ClasspathElementFileDir extends ClasspathElement {
             private final Runnable onClose = new Runnable() {
                 @Override
                 public void run() {
-                    if (isOpen.get()) {
-                        close();
-                    }
+                    close();
                 }
             };
 
@@ -298,8 +296,10 @@ class ClasspathElementFileDir extends ClasspathElement {
                         nestedJarHandler.markSliceAsClosed(fileSlice);
                         fileSlice = null;
                     }
+
+                    // Close inputStream
+                    super.close();
                 }
-                super.close(); // Close inputStream
             }
         };
     }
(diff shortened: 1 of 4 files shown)


lukehutch:
LGTM, thanks, good thinking.

Right, I plan to drop JDK 7 support in ClassGraph version 5 (which I will probably work on after there have been no new bug reports for a few months, which the project is approaching). At that point I'll switch the code internally to use lambdas, method references, etc.

For now though, with your change in place, I'll probably just pass a reference in to the `Resource` to close, rather than a `Runnable` that just calls `Resource#close()`.

chrisr3:
OK, thanks. BTW ClassGraph 4.8.135 hasn't appeared in [Maven Central](https://repo.maven.apache.org/maven2/io/github/classgraph/classgraph/) yet. Was there a problem during publication?

Out of curiosity, will ClassGraph 5 still support JDK8 please?

lukehutch:
Yes, the plan is to bump the minimum supported JDK version from 7 to 8 for ClassGraph version 5. JDK 7 support has required quite a lot of workarounds over the years. But JDK 8 will still be important for probably another 5-10 years or something, so I don't want to increase the minimum supported version beyond JDK 8 for a long time.

I made the changes, you can look at my commit 220be7b if you are curious. However this does not change functionality in any way, so I won't push out a new release just for this.

Thanks again for your code contributions and your ideas!

lukehutch:
> OK, thanks. BTW ClassGraph 4.8.135 hasn't appeared in [Maven Central](https://repo.maven.apache.org/maven2/io/github/classgraph/classgraph/) yet. Was there a problem during publication?

I'm not sure what the problem is -- the Sonatype Nexus UI is just hanging, so I suspect Sonatype is broken, and I use Sonatype to publish artifacts to Maven Central. (This sort of thing happens a lot with Sonatype...)

https://oss.sonatype.org/#nexus-search;quick~io.github.classgraph

lukehutch:
@chrisr3

(The Sonatype UI is now responding, a few minutes later. I see that 4.8.135 is not there. I have no idea what happened! I re-tagged this release, including with the latest refactoring changes, and published this release again.)

Now 4.8.135 is showing up in Sonatype, so it should be visible in Maven Central within a few hours.

Issue #633, opened 2022-01-13
Thrown exceptions on `MethodInfo`


jkschneider:
Hi 👋 

I'd love to add thrown exceptions support to `MethodInfo`. Thrown exceptions show up on `getTypeSignatureOrTypeDescriptor().getThrowsSignature()` apparently only when at least one of the exceptions is a generic type variable.

There appears to be a particular `attributeNameCpIdx` named `Exceptions` that has the data for all other method signatures with exceptions. If you're willing, I am happy to provide a PR.

We're using classgraph quite heavily in https://github.com/openrewrite/rewrite. Thanks!

lukehutch:
I'm glad you're getting some good use out of ClassGraph! I hadn't heard of Rewrite, it looks like an awesome project.

That would be awesome, if you'd be willing to contribute this! Thanks for the offer.

Here are the pointers to what you will need to do (roughly... YMMV):

Pick up the `Exceptions` attribute roughly here:

https://github.com/classgraph/classgraph/blob/latest/src/main/java/io/github/classgraph/Classfile.java#L1655

Plumb it through to `MethodInfo`:

https://github.com/classgraph/classgraph/blob/latest/src/main/java/io/github/classgraph/Classfile.java#L1677

Note that ClassGraph goes out of its way to only lazy-load information, so don't try parsing type descriptors or try to look up classes by name, leave them in string form, and parse or look them up as needed, caching the result, inside of `MethodInfo` (see what has been done for `MethodInfo.getTypeDescriptor` etc.)

You need to index the returned exception class names in `MethodInfo.findReferencedClassInfo`:

https://github.com/classgraph/classgraph/blob/latest/src/main/java/io/github/classgraph/MethodInfo.java#L885

If you do parse any type signatures or look up / create any `ClassInfo` objects (lazily) in `MethodInfo`, then ensure that the `ScanResult` is set in those reachable objects once they are created -- see an example here:

https://github.com/classgraph/classgraph/blob/latest/src/main/java/io/github/classgraph/MethodInfo.java#L254

and then to be safe, also add code here to set the `ScanResult` in those cached objects if they are non-null, at class graph creation time:

https://github.com/classgraph/classgraph/blob/latest/src/main/java/io/github/classgraph/MethodInfo.java#L847

Also add the `throws` notation to the `toString` method, avoiding doubling-up exceptions, if there's any chance that exceptions will be listed twice, once in `getThrowsSignatures()` and once in the `Exceptions` attribute:

[...]

PR #637, opened 2022-01-14
Add thrown exceptions to MethodInfo (#633)

jkschneider:


src/main/java/io/github/classgraph/MethodInfo.java
```diff
@@ -318,6 +324,24 @@ public String getTypeSignatureOrTypeDescriptorStr() {
         }
     }
 
+    public ClassInfoList getThrownExceptions() {
```
jkschneider, line 327: When a thrown exception is a generic type variable, the `Exceptions` attribute refers to the bound type, which I think makes sense to return here?

lukehutch: Yes, I think so, or if all else fails and the type variable cannot be resolved or does not have a supertype (e.g. `E extends IOException`), just return `Exception` as the concrete type, I think.

It's a shame Java doesn't have union types, since it would be nice to return `TypeVariable` or `ClassInfo` depending on whether or not the class variable is unresolved.

Also you can discard type parameters like `T` in `MyException<T>`. If anyone needs those, they can use the existing method to get the generic `throws` type signature.

src/main/java/io/github/classgraph/MethodInfo.java
```diff
@@ -318,6 +324,24 @@ public String getTypeSignatureOrTypeDescriptorStr() {
...
+        if (thrownExceptions == null && thrownExceptionNames != null) {
+            thrownExceptions = new ClassInfoList(thrownExceptionNames.length);
+            for (String thrownExceptionName : thrownExceptionNames) {
+                ClassInfo classInfo = scanResult.getClassInfo(thrownExceptionName);
+                thrownExceptions.add(classInfo);
```
lukehutch: `classInfo` may be `null` here -- is it helpful to have a `null` placeholder in the list for classes that could not be found (because they were not discovered during the scan)?

Actually I forgot one more change you should make -- look at `extendScanningUpwards()` -- you should schedule the thrown exceptions for scanning in that method. If you make this change, the chance of `classInfo` being `null` is lower, because any named exceptions will be scanned as "external classes" even if they are not in the accept list.

That raises the question though, is the `Exception` attribute always present, even if there is a generic `throws` type signature present?

src/main/java/io/github/classgraph/MethodInfo.java
jkschneider: > Also you can discard type parameters like T in MyException<T>

I believe that is how this functions currently.

src/main/java/io/github/classgraph/MethodInfo.java
jkschneider: > That raises the question though, is the Exception attribute always present, even if there is a generic throws type signature present?

Yes it is. The `Exception` attribute will return the bound type though. So it just kind of works as is without having to strip parameterizations and such off of throws.

jkschneider: > is it helpful to have a null placeholder in the list for classes

Null values removed in the latest commit.

lukehutch:
Looks good to me. Thanks for all your work on this! Will merge now.

lukehutch:
I'm working on another bugfix, which might take a couple of days to sort out, but I'll push this out in a new version as soon as I can. Thanks for your patience, I'm swamped right now!

jkschneider:
> Thanks for your patience, I'm swamped right now!

No worries, and thanks for your incredible responsiveness.

Issue #641, opened 2022-01-31
ClassGraph fails to discover resources and classes in Quarkus 2.7 during build time


michael-simons:
First of all, I am super sorry for that really awful discussion over here https://github.com/quarkusio/quarkus/discussions/23312 in case you got swamped with notifications already. I didn't expect that.

As usually, I am using ClassGraph to find both resources and classes. I do this in Quarkus build time in the extension in question. 

Working combination
Quarkus 2.6.2 plus ClassGraph 4.8.138

Broken combination
Quarkus 2.6.2 plus ClassGraph 4.8.138

JDK 17.

Here's the description from the discussion above. Down below I added the verbose log for both working and broken:

I maintain Neo4j-Migrations: https://github.com/michael-simons/neo4j-migrations, which let's you execute Cypher scripts (resources) and Java classes against Neo4j.

*I use ClassGraph to discover those resources.*
Since last release I do support Quarkus in native mode. To make this possible, I moved all discovering of things into build time / augmentation. That works well with Quarkus *2.6.x* (all releases).

Those classes:

https://github.com/michael-simons/neo4j-migrations/blob/0db309476f9c737b1b4674193f838f7e3c74ec75/neo4j-migrations-quarkus-parent/runtime/src/main/java/ac/simons/neo4j/migrations/quarkus/runtime/StaticClasspathResourceScanner.java#L57-L70

and

https://github.com/michael-simons/neo4j-migrations/blob/0db309476f9c737b1b4674193f838f7e3c74ec75/neo4j-migrations-quarkus-parent/runtime/src/main/java/ac/simons/neo4j/migrations/quarkus/runtime/StaticJavaBasedMigrationDiscoverer.java#L48-L61

are invoked from a build time processor, both in 2 steps (creating the scanner, than executing them and adding things to the augmented classes), like this:

https://github.com/michael-simons/neo4j-migrations/blob/0db309476f9c737b1b4674193f838f7e3c74ec75/neo4j-migrations-quarkus-parent/deployment/src/main/java/ac/simons/neo4j/migrations/quarkus/deployment/MigrationsProcessor.java#L58-L83

With Quarkus 2.7.x this suddenly fails: https://github.com/michael-simons/neo4j-migrations/runs/5002711436?check_suite_focus=true#step:9:1522 no resources are found anymore.

What fails exactly is

```java
try (var scanResult = new ClassGraph().acceptPaths(paths).scan()) {
[...]

PR #642, opened 2022-01-31
Fix QuarkusClassLoaderHandler on Quarkus 2.7.

michael-simons:
In Quarkus 2.7, the class loader stopped returning `JarClassPathElement`
or `DirectoryClassPathElement` entries. Instead, some implementation of
`ClassPathElement` is returned. This change tries to access it's
`getRoot` method, returning a `java.nio.file.Path` pointing to the root
of the element, as supported with class graph.

The behaviour has been verified on Quarkus 2.6 and Quarkus 2.7.

This is a fix for #641, thanks @lukehutch in advance for considering this input. It would help Neo4j a lot.

### DIFF
--- a/src/main/java/nonapi/io/github/classgraph/classloaderhandler/QuarkusClassLoaderHandler.java
+++ b/src/main/java/nonapi/io/github/classgraph/classloaderhandler/QuarkusClassLoaderHandler.java
@@ -127,6 +127,13 @@ class QuarkusClassLoaderHandler implements ClassLoaderHandler {
             } else if ("io.quarkus.bootstrap.classloading.DirectoryClassPathElement".equals(elementClassName)) {
                 classpathOrder.addClasspathEntry(ReflectionUtils.getFieldVal(false, element, "root"), classLoader,
                         scanSpec, log);
+            } else {
+                Object rootPath = ReflectionUtils.invokeMethod(false, element, "getRoot");
+                if (rootPath instanceof Path) {
+                    classpathOrder.addClasspathEntry(ReflectionUtils.invokeMethod(false, element, "getRoot"),
+                        classLoader,
+                        scanSpec, log);
+                }
             }
         }
     }


michael-simons:
Do you need anything more / help from me?

src/main/java/nonapi/io/github/classgraph/classloaderhandler/QuarkusClassLoaderHandler.java
```diff
@@ -127,6 +127,13 @@ private static void findClasspathOrderForQuarkusClassloader(final ClassLoader cl
             } else if ("io.quarkus.bootstrap.classloading.DirectoryClassPathElement".equals(elementClassName)) {
                 classpathOrder.addClasspathEntry(ReflectionUtils.getFieldVal(false, element, "root"), classLoader,
                         scanSpec, log);
+            } else {
+                Object rootPath = ReflectionUtils.invokeMethod(false, element, "getRoot");
+                if (rootPath instanceof Path) {
+                    classpathOrder.addClasspathEntry(ReflectionUtils.invokeMethod(false, element, "getRoot"),
```
lukehutch, line 133: I'll merge this, but why did you invoke this method twice? Why not just use the value of `rootPath`?

lukehutch:
> Do you need anything more / help from me?

No, sorry, I have just been traveling for the last 3 weeks, with very limited time... Merging this now -- thank you for the contribution!

src/main/java/nonapi/io/github/classgraph/classloaderhandler/QuarkusClassLoaderHandler.java
michael-simons: Ooops. That was an oversight. You fixed this, didn't you? Thank you for that!

lukehutch: I did, no problem!

lukehutch:
@michael-simons I had a bunch of other bugfixes I wanted to get in before the next release, but since I'm finding very little time while traveling, and since this seems like an urgent bug for your team, I will push a release out now. Should be fixed in 4.8.139. Thanks for the patch!

Issue #651, opened 2022-02-15
NPE in JBossClassLoaderHandler#findClasspathOrder 


arthware:
In our scenario we get a `NullPointerException` in `JBossClassLoaderHandler#findClasspathOrder `

The `moduleMap` is null in our case, leading to an NPE in the next line, when accessing the entrySet.

https://github.com/classgraph/classgraph/blob/f874380e58ac3e9e4834df199129a58c79d15f33/src/main/java/nonapi/io/github/classgraph/classloaderhandler/JBossClassLoaderHandler.java#L213-L215

I was able to fix this with a simple null check. The scan returns the expected results. 

**Before:** 
` for (final Entry<Object, Object> ent : moduleMap.entrySet() ) {`

**After:**
` Set<Entry<Object, Object>> moduleMapEntries = moduleMap != null ? moduleMap.entrySet() : Collections.emptySet();
    for (final Entry<Object, Object> ent : moduleMapEntries) {`

lukehutch:
Fixed via #652. Thank you!

arthware:
Thank you for merging and your efforts!

arthware:
@lukehutch any chance to get a release pushed out in the next two weeks? That'd be awesome. Thank you very much!

lukehutch:
Yes, I'm sorry for the delay. I will do it in a few hours once my current flight lands!

lukehutch:
OK, I was able to push out version 4.8.140 with this fix. I probably should have just done this after merging. Thanks again for your fix!

PR #652, opened 2022-02-15
#651 fixes NPE in JbossClassLoaderHandler

arthware:


lukehutch:
Thank you for this fix! Looks good to me. I'm traveling right now and have a few other things to fix, so I'm not sure when I can push out a new version.

PR #707, opened 2022-10-09
Add ClassMemberInfo as super class of FieldInfo and MethodInfo

sebthom:
This PR adds a ClassMemberInfo class that holds shared fields/methods of FieldInfo/MethodInfo classes.


lukehutch:
@sebthom Thanks for this great contribution, and sorry for the delay in merging this! I don't know why it didn't click for me before that `FieldInfo` and `MethodInfo` were so similar.

PR #732, opened 2022-11-21
Update javadoc

sebthom:
Replace usage of nouns "field", "method" with "class member"


lukehutch:
@sebthom Thanks for fixing the Javadoc!

PR #740, opened 2022-12-19
Add `ClassInfo#getSourceFile`

freya022:
This PR adds the `SourceFile` attribute, as specified [in the 4.7.10 section](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.10)

### DIFF
--- a/src/main/java/io/github/classgraph/ClassInfo.java
+++ b/src/main/java/io/github/classgraph/ClassInfo.java
@@ -2934,9 +2934,11 @@ public class ClassInfo extends ScanResultObject implements Comparable<ClassInfo>
 
     /**
      * Returns the name of the source file this class has been compiled from,
-     * such as {@code ClassInfo.java} or {@code KClass.kt}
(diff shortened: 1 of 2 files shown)


freya022:
Btw this would be nice if we could use some nullability annotations, as some fields (such as this one) are optional, letting IDEs or even languages (like Kotlin) to know what is nullable or not, would be useful

I'd suggest using this one: `compileOnly("com.google.code.findbugs:jsr305:3.0.2")` as a compile only dependency as to not be included in the uber jar

lukehutch:
Thank you for the contribution, @freya022! I'll roll out a release tonight.

Good suggestion re findbugs. I did some nullability dataflow analysis in the past, and I gave up, partly because the compile-time dependency I added wasn't compatible with JDK 7, which was (and is still) supported by ClassGraph. I'll try out findbugs.

I suspect it will be an enormous amount of work to get the nullability annotations correct throughout the entire project, however. I am not up to date in my knowledge about how much of the work tools can do correctly for the programmer.

PR #742, opened 2022-12-20
Add `getTypeAnnotationInfo` to `TypeParameter` and `TypeArgument`

### LINKED ISSUE #741: Missing API access for annotations on type parameters and wildcard type arguments
`TypeParameter` and `TypeArgument` (in case of using a wildcard) could hold annotations, but they do not have a `getTypeAnnotationInfo()` getter exposed, despite parsing works seemingly correctly in both cases. 

Consider analysing annotations specified on the method parameter types in the scan result of the following class:

```java
class GenericTypeAnnotationTest {
    @Target({ ElementType.FIELD, ElementType.TYPE_USE, ElementType.TYPE_PARAMETER })
    private static @interface A {
    }

    <@A T> void saveT(T param) {
    }

    void saveUList(List<@A ?> param) {
    }
}
```

`typeParameter.toString()` and `typeArgument.toString()` results show the `@A` annotation where expected, but there is no API to access the `@A` annotation info other than `toString()`.

Apart from Java, this has a big impact on working with Kotlin compiled code, as for some types (`BigDecimal` and `BigInteger` are known examples, my wild guess, for supporting Kotlin language extensions) compiler emits `List<@A ? extends Type> param` in JVM bytecode whereas the source had `List<@A Type> param`.

platosha:
Fixes classgraph/classgraph#741

Exposes missing API for accessing parsed annotations on type parameters and wildcard type arguments.

### DIFF
--- a/src/main/java/io/github/classgraph/TypeArgument.java
+++ b/src/main/java/io/github/classgraph/TypeArgument.java
@@ -225,6 +225,15 @@ public final class TypeArgument extends HierarchicalTypeSignature {
         }
     }
 
+    /**
+     * Get a list of {@link AnnotationInfo} objects for any type annotations before the type argument wildcard, or null if none.
+     * 
+     * @return a list of {@link AnnotationInfo} objects for any type annotations before the type argument wildcard, or null if none.
+     */
+    public AnnotationInfoList getTypeAnnotationInfo() {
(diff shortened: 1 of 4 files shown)


lukehutch:
@platosha thank you so much for the contribution, including the unit test!

PR #743, opened 2022-12-22
Fixed typo

cowwoc:


lukehutch:
Thank you for the pull request, @cowwoc! I won't push out a release for this, but it will be in the next release whenever that happens.

Issue #772, opened 2023-05-12
ClassInfo#getMethodInfo picks wrong inherited method


peter-schmitz:
Hi,

I believe ClassInfo#getOverrideOrder() is incorrect for methods and should probably prefer methodInfos from direct super classes before any interface methods.
After a quick test, the current order does appear to be correct for fields / interface constants.

Tested with JDK 8 and JDK 17.

Example:

Class: java.io.DataOutputStream
Method "close":

`ClassLoader.getSystemClassLoader().loadClass("java.io.DataOutputStream").getMethod("close")`

-> Returns: `public void java.io.FilterOutputStream.close() throws java.io.IOException`

```
ClassGraph graph = new ClassGraph()
        .enableClassInfo()
        .enableMethodInfo()
        .acceptClasses("java.io.DataOutputStream")
        .enableSystemJarsAndModules();

try (ScanResult scan = graph.scan()) {
    ClassInfo classInfo = scan.getClassInfo("java.io.DataOutputStream");
    MethodInfoList closeMethods = classInfo.getMethodInfo("close");
    for (MethodInfo closeMethod : closeMethods) {
        System.out.println(closeMethod + " from " + closeMethod.getClassInfo().getName());
    }
    //For some reason, the classInfo.getMethodInfo() list contains the abstract close method as well?
}
```

prints

`public abstract void close() throws java.io.IOException from java.io.Closeable`

Expected output:
`public void close() throws java.io.IOException from java.io.FilterOutputStream`

lukehutch:
Thanks for the bug report. I took a guess at what the correct order would be, but you're right, it should be changed.

Do you have time to submit a pull request?

peter-schmitz:
Sorry, this week is quite full for me already, so probably next week at the earliest :/

peter-schmitz:
Done, see #773

lukehutch:
@peter-schmitz Many thanks for fixing this! Merged and released as 4.8.159.

lukehutch:
Actually make that 4.8.160, `maven-release-plugin` was playing up, and I don't think the changes went into 4.8.159...

PR #773, opened 2023-05-23
#772 Fix method override order

peter-schmitz:
Fixes order of classes overriding one another for methods, preventing getMethodInfo from returning a method that is overridden.


lukehutch:
Looks good to me. Thanks for doing the hard work to fix this!

Issue #780, opened 2023-07-05
getResourcesWithPath can return null


corebonts:
Since [this change](https://github.com/classgraph/classgraph/commit/cd213efa8490cbe6862ef1a8f3debd461a6f5603#diff-ca7988d4cc5346f9d77b470beda2e1f9021f8823de01c99d349c42c1ef8141d0R515) getResourcesWithPath() can return null if it's called n-th times for a non-scanned or non-existing path.

```
    @Test
    public void getResourcesWithPathShouldNeverReturnNull() {
        try (ScanResult result = new ClassGraph().scan()) {
            for (int i = 0; i < 10; i++) {
                assertThat(result.getResourcesWithPath("/some/non/existing/path")).isNotNull().isEmpty();
            }
        }
    }
```

lukehutch:
Fixed and released in 4.8.161. Thank you for the bug report and the fix!

PR #781, opened 2023-07-05
Never return null in getResourcesWithPath() #780

corebonts:
Since cd213ef getResourcesWithPath() returned null if it's called n-th times for a non-scanned or non-existing path.


lukehutch:
LGTM, thanks for the fix! I'll push out a release right now with the fix.

Issue #787, opened 2023-07-10
Get all versions of a class in a Multi-Release JAR


pkesseli:
When invoking ClassGraph as follows:

```java
final ClassGraph classGraph = new ClassGraph().disableNestedJarScanning().ignoreClassVisibility();
try (final ScanResult scanResult = classGraph.verbose().scan()) {
    final ResourceList classes = scanResult.getAllResources();
    return classes.parallelStream().filter(resource -> resource.getPath().contains("StackLocator.class"))
            .map(resource -> resource.getPathRelativeToClasspathElement()).collect(Collectors.toList());
}
```

I only get the class `META-INF/versions/9/org/apache/logging/log4j/util/StackLocator.class`, not the base version `/org/apache/logging/log4j/util/StackLocator.class`. I would like to instrument all versions of the class, so how do I configure ClassGraph to iterate over all versions of classes?

Attached a minimal Maven project to reproduce the issue as well as the verbose ClassGraph output:
[classgraph-demo.tar.gz](https://github.com/classgraph/classgraph/files/12005752/classgraph-demo.tar.gz)
[classgraph.log](https://github.com/classgraph/classgraph/files/12005754/classgraph.log)

lukehutch:
Hi, this is interesting -- by design, only a single classfile with a given package path (specifically the one that would be found in the current runtime environment by the running JVM) is returned by ClassGraph, when scanning classes. However, you are supposed to be able to find all versions of a resource, as your code attempts to do.

I made the assumption when I wrote the code that resources should also follow the `META-INF/versions` rules in the current runtime environment, if there are multiple versions available. The thinking here was that you should only be able to see one version of a resource *per package path, per classpath element*.

Before JVM 9, there could only be one version of a resource per package path, because the package path was defined relative to the root of the classpath element. Classfile versioning changed this, so you can have multiple versions of a resource within a single classpath element -- one per JVM version number, and one unversioned (legacy classpath) version.

Finding only the version that is relevant to the current runtime environment (the current behavior) is going to be the right thing to do for 99% of users. (Actually even the Java module system in the running JVM will not show you more versions of a resource than the one that works with the current JVM version...) But what we'll need to do for your usecase is disable version shadowing entirely.

I'm really tight on time right now -- do you have time to look into how to do this?

[...]

PR #788, opened 2023-08-06
#787 Get all versions in multi-release JAR

pkesseli:
Allow retrieving all resource versions in a multi-release JAR.

### DIFF
--- a/src/main/java/io/github/classgraph/ClassGraph.java
+++ b/src/main/java/io/github/classgraph/ClassGraph.java
@@ -207,12 +207,14 @@ public class ClassGraph {
     }
 
     /**
-     * Enables the scanning of classfiles, producing {@link ClassInfo} objects in the {@link ScanResult}.
+     * Enables the scanning of classfiles, producing {@link ClassInfo} objects in the {@link ScanResult}. Implicitly
+     * disables {@link #enableMultiReleaseVersions()}.
      *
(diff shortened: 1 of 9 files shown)


src/main/java/io/github/classgraph/ClassGraph.java
```diff
@@ -1429,6 +1431,31 @@ public ClassGraph enableMemoryMapping() {
...
+     * If true, provide all versions of a multi-release resource using their multi-release path prefix, instead of
+     * just the one the running JVM would select. Implicitly disables {@link #enableClassInfo()} and all features
+     * depending on it.
+     * 
+     * @return this (for method chaining).
+     */
+    public ClassGraph enableMultiReleaseVersions() {
+        scanSpec.enableMultiReleaseVersions = true;
+
+        scanSpec.enableClassInfo = false;
```
pkesseli, line 1444: I wasn't sure whether this was the idiomatic approach or whether `enableMultiReleaseVersions()` should just throw an `IllegalArgumentException` if `enableClassInfo()` was set and vice-versa.

lukehutch: I think this behavior is fine, given that you documented it.

lukehutch:
Looks good to me. Thank you for the contribution!!

PR #793, opened 2023-09-08
Fix patched module resource lookup

### LINKED ISSUE #704: Duplicate resources found with Maven / Surefire patching test resources into the module path.
Find a simple reproducer attached ([mpcg.zip](https://github.com/classgraph/classgraph/files/9623674/mpcg.zip)): Maven / IntelliJ and friends will patch test classes and test resources into the main module when there's a `module-info.java` in the project (`--patch-module mpcg=/Users/msimons/Projects/misc/mpcg/target/test-classes -`). 

When this happens, classgraph will discover resources multiple times. 

I don't repeat the code from the reproducer here, 

```
.
├── pom.xml
└── src
    ├── main
    │   ├── java
    │   │   ├── foo
    │   │   │   └── App.java
    │   │   └── module-info.java
    │   └── resources
    └── test
        ├── java
        │   └── foo
        │       └── AppTest.java
        └── resources
            └── stuff
                └── whatever.cypher

10 directories, 5 files
```

In this scenario, 

```
ScanResult scanResult = new ClassGraph().acceptPaths("stuff").scan();
ResourceList allResources = scanResult.getAllResources().nonClassFilesOnly()
```

will give me `whatever.cypher` twice

```
[file:/Users/msimons/Projects/misc/mpcg/target/classes/stuff/whatever.cypher,
    file:/Users/msimons/Projects/misc/mpcg/target/test-classes/stuff/whatever.cypher]
```

This can be fixed by

```
<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-surefire-plugin</artifactId>
				<version>3.0.0-M7</version>
				<configuration combine.self="append">
					<!-- Add this and the test works… -->
					<useModulePath>false</useModulePath>
				</configuration>
			</plugin>
```

but this is maybe not desired to skip the module path completely.

I am unsure if its an Maven issue or Classpath issue, though.

cc @sbrannen you expressed interest in this on twitter.

Sineaggi:
The current getURI method constructs a classpath resource from the `getClasspathElementURI` method. This doesn't worked for patched modules that could come from different classpaths (for example the module may start as a jar but have a directory patched in).

Calling the `ModuleReader#find` method (https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/module/ModuleReader.html#find-java.lang.String-) instead will give us the URI of the resource.

Might fix https://github.com/classgraph/classgraph/issues/704

### DIFF
--- a/src/main/java/io/github/classgraph/ClasspathElementModule.java
+++ b/src/main/java/io/github/classgraph/ClasspathElementModule.java
@@ -184,6 +184,20 @@ class ClasspathElementModule extends ClasspathElement {
             }
 
             @Override
+            public URI getURI() {
+                try {
+                    ModuleReaderProxy localModuleReaderProxy = moduleReaderProxyRecycler.acquire();
+                    try {
+                        return localModuleReaderProxy.find(resourcePath);
+                    } finally {
+                        moduleReaderProxyRecycler.recycle(localModuleReaderProxy);
+                    }
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+            }
+
+            @Override
             public InputStream open() throws IOException {
                 if (skipClasspathElement) {
                     // Shouldn't happen
--- a/src/main/java/io/github/classgraph/ModuleReaderProxy.java
+++ b/src/main/java/io/github/classgraph/ModuleReaderProxy.java
@@ -31,6 +31,7 @@ package io.github.classgraph;
 import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URI;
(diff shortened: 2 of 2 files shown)


Sineaggi:
Hmm, on further investigation perhaps it's the `getClasspathElementURI` that needs fixing.

lukehutch:
This is a really high-quality PR -- thanks for the work on this and the attention to detail as to how code works in ClassGraph! However, based on your last comment, I'll let you dig a bit further before I merge this. (Sorry, I'm really short on time to help out with this right now...) Please let me know what you figure out.

PR #850, opened 2024-04-17
Reduce scan time by avoid having redundant file access

attilapuskas:
These changes are meant to reduce the general file access and keep the same scan results as before.
Most of the time it is about removing some duplicated _canRead_, _isFile_ and similar calls that are usually top contibutors of the scanning time especially on Windows based systems with any kind of antivirus/firewall enabled.

I have tested these on a project using ~500 classpath entries, from which ~380 are jar files. Thanks to the extended ClassGraph configuration it was possible to filter out most of the artifacts that are not relevant, but even in that state it was a measurable improvement.
| | before | changes included |
|  :---: |  :---: | :---: |
| all | 21 s | 11 s  |
| filtered |  9 s | 6 s  |

### DIFF
--- a/src/main/java/io/github/classgraph/ClasspathElementDir.java
+++ b/src/main/java/io/github/classgraph/ClasspathElementDir.java
@@ -111,17 +111,20 @@ class ClasspathElementDir extends ClasspathElement {
                 final Path libDirPath = classpathEltPath.resolve(libDirPrefix);
                 if (FileUtils.canReadAndIsDir(libDirPath)) {
                     // Add all jarfiles within the lib dir as child classpath entries
-                    try (DirectoryStream<Path> stream = Files.newDirectoryStream(libDirPath)) {
+                    try (DirectoryStream<Path> stream = Files.newDirectoryStream(libDirPath, new DirectoryStream.Filter<Path>() {
+                        @Override
+                        public boolean accept(Path filePath) {
+                            return filePath.toString().toLowerCase().endsWith(".jar") && Files.isRegularFile(filePath);
+                        }
+                    })) {
                         for (final Path filePath : stream) {
-                            if (Files.isRegularFile(filePath) && filePath.toString().toLowerCase().endsWith(".jar")) {
-                                if (log != null) {
-                                    log(classpathElementIdx, "Found lib jar: " + filePath, log);
-                                }
-                                workQueue.addWorkUnit(new ClasspathEntryWorkUnit(filePath, getClassLoader(),
-                                        /* parentClasspathElement = */ this,
-                                        /* orderWithinParentClasspathElement = */ childClasspathEntryIdx++,
-                                        /* packageRootPrefix = */ ""));
+                            if (log != null) {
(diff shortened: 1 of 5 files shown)


lukehutch:
@attilapuskas Looks good to me -- this is an epic PR! Thank you!

lukehutch:
Released in 4.8.171. Many thanks!

attilapuskas:
Thanks for the very quick response! :)

src/main/java/io/github/classgraph/ClasspathElementDir.java
```diff
@@ -396,16 +401,19 @@ private void scanPathRecursively(final Path path, final LogNode log) {
...
+        FileUtils.FileAttributesGetter getFileAttributes = FileUtils.createCachedAttributesGetter();
 
         // Determine whether this is a modular jar running under JRE 9+
         final boolean isModularJar = VersionFinder.JAVA_MAJOR_VERSION >= 9 && getModuleName() != null;
 
         // Only scan files in directory if directory is not only an ancestor of an accepted path
         if (parentMatchStatus != ScanSpecPathMatch.ANCESTOR_OF_ACCEPTED_PATH) {
             // Do preorder traversal (files in dir, then subdirs), to reduce filesystem cache misses
-            for (final Path subPath : pathsInDir) {
+            for (final Path subPath : new ArrayList<>(pathsInDir)) {
```
szarnekow, line 412: I think you could avoid the copy of the list and the subsequent linear scans for `pathsInDir.remove` by using an Iterator instead of the for-loop along with `Iterator.remove`.

src/main/java/io/github/classgraph/ClasspathElementDir.java
```diff
@@ -396,16 +401,19 @@ private void scanPathRecursively(final Path path, final LogNode log) {
             return;
         }
         Collections.sort(pathsInDir);
+        FileUtils.FileAttributesGetter getFileAttributes = FileUtils.createCachedAttributesGetter();
```
szarnekow, line 404: Can you help me understand the purpose of the cached attributes here? As far as I see it, each path is used once with the getFileAttributes. I'm likely missing smth.

src/main/java/io/github/classgraph/ClasspathElementDir.java
attilapuskas: Indeed, I just selected one of the smallest & easy change to avoid any side effects and considered that one also rather cheap, at least negligible regarding the whole performance.

src/main/java/io/github/classgraph/ClasspathElementDir.java
attilapuskas: Nice catch! I believe when I introduced the cache based lookup it was still the case that we needed it. Maybe I had a version were _'Recurse into subdirectories'_ also used the already cached attributes to check if a path is a directory and I missed it from the final state.

lukehutch: Feel free to take another pass at this! Thanks!

PR #851, opened 2024-04-19
ClasspathElementDir review adjustments

attilapuskas:
Based on the following comments:
https://github.com/classgraph/classgraph/pull/850#discussion_r1570337208
https://github.com/classgraph/classgraph/pull/850#discussion_r1570348184

### DIFF
--- a/src/main/java/io/github/classgraph/ClasspathElementDir.java
+++ b/src/main/java/io/github/classgraph/ClasspathElementDir.java
@@ -43,6 +43,7 @@ import java.nio.file.attribute.PosixFilePermission;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


lukehutch:
LGTM. Thanks!

lukehutch:
Released in 4.8.172.

attilapuskas:
Thanks!
Do you know by any chance why build (windows-latest, 15 & 18) have failed? They give the results in different order, but I have no idea how the commits could cause such a thing.

lukehutch:
I opened a bug about that. There's some nondeterminism in the classpath resolution order which is not supposed to be there (the order is supposed to be deterministic, despite parallel scanning). I don't know what could cause that, and it's going to be a hard bug to get to the bottom of. I don't have the time to look into it right now. But if you look at the failing test, you might be able to figure it out.

The issue occurs when a classpath element adds additional classpath elements to the scan, partway through the scan. Specifically, when a manifest has a Class-Path entry.

attilapuskas:
I see, thanks for the detailed explanation!

PR #869, opened 2024-06-18
Issues/843 jboss vfs adjustments

dominik2611:


lukehutch:
Thank you @dominik2611!

I don't have a way to verify this is complete and correct, since I don't know anything about this runtime environment, but the code looks good to me.

PR #876, opened 2024-07-26
Add a test for Issue 875

cushon:
This may be working as intended. It looks like on JDK 9+, the scan is including `java.class.path`, I think because of the `moduleFinder != null && moduleFinder.forceScanJavaClassPath())` case here:

https://github.com/classgraph/classgraph/blob/8bfaa5776f22ca45ff8cc1c552d55510e754b2dd/src/main/java/nonapi/io/github/classgraph/classpath/ClasspathFinder.java#L299-L302

The test passes on JDK 11:

```
JAVA_HOME=/Library/Java/JavaVirtualMachines/zulu-11.jdk/Contents/Home/ ./mvnw test -Dtest=io.github.classgraph.issues.issue875.Issue875Test
...
INFO: 2024-07-26T10:03:31.542-0700  ClassGraph  Getting classpath entries from java.class.path
```

and fails on JDK 8

```
JAVA_HOME=/Library/Java/JavaVirtualMachines/zulu-8.jdk/Contents/Home/ ./mvnw test -Dtest=io.github.classgraph.issues.issue875.Issue875Test
...
INFO: 2024-07-26T10:02:23.108-0700  ClassGraph  External superclass io.github.classgraph.issues.issue875.Issue875Test$SuperCls was not found in non-rejected packages -- cannot extend scanning to this class
...
Expecting ArrayList:
  []
to contain only:
  ["io.github.classgraph.issues.issue875.Issue875Test$SuperCls",
    "io.github.classgraph.issues.issue875.Issue875Test$Cls"]
```

### DIFF
--- a/src/test/java/io/github/classgraph/issues/issue875/Issue875Test.java
+++ b/src/test/java/io/github/classgraph/issues/issue875/Issue875Test.java
@@ -1,0 +1,103 @@
+/*
+ * This file is part of ClassGraph.
+ *
+ * Author: Luke Hutchison
+ *
+ * Hosted at: https://github.com/classgraph/classgraph
+ *
+ * --
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2019 Luke Hutchison
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+ * documentation files (the "Software"), to deal in the Software without restriction, including without
+ * limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so, subject to the following
+ * conditions:
+ *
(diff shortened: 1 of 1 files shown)


lukehutch:
Super interesting test-case, thanks! I assume what is happening on JDK 11 is that the `SuperSuperCls` class and its descendants are actually being scanned from the module path, not the classpath.

When you scan classes from the traditional classpath, at most one superclass-to-subclass relationship is stored in the `ScanResult` (this is possible because a class definition specifies its immediate superclass). ClassGraph does try to add the superclass and super-super-class in to the scan, by following the parent class relationship upwards, but in this case, you don't even include those class defs in the temporary jarfile, so it can't find them.

Unless, that is, these classes were already scanned from the module path. Then the classpath version is just a duplicate, and can be ignored.

So yes, this is working as intended, I believe...

PR #878, opened 2024-08-17
Store annotation getters of ClassInfo and MethodInfo

freya022:
As those are the most likely used, this could save considerable CPU time

In my workload these changes produce a 2x improvement (~167ms -> ~84ms), as I run over some classes multiple times

### DIFF
--- a/src/main/java/io/github/classgraph/ClassInfo.java
+++ b/src/main/java/io/github/classgraph/ClassInfo.java
@@ -179,6 +179,12 @@ public class ClassInfo extends ScanResultObject implements Comparable<ClassInfo>
      */
     private transient List<ClassInfo> methodOverrideOrder;
 
+    /** The annotations, once they are loaded */
+    private ClassInfoList annotationsRef;
+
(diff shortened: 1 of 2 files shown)


lukehutch:
@freya022 Thanks for the contribution! I didn't realize these same methods would be called repeatedly on the same classes, so I never cached them. Are you duplicating work in your code?

There are probably some other method calls that could be memoized in this way too, but I don't have the time to review/audit the API surface right now.

The downside of this is of course greater memory overhead while the `ScanResult` is still in scope and not closed. People already complain about `ScanResult` producing an enormous object graph that can cause OOM on low-memory systems. But since you got such a big speedup from this change, it is probably worth it to cache this.

freya022:
Yes, there are multiple places where i need to check for the annotations, a workaround would be to create my own objects which store the CG object and a few other properties I often need, having the annotations stored on the CG object is cleaner and performant enough

lukehutch:
Released in 4.8.175. Thanks!

Issue #810, opened 2023-11-16
Issue 673 test sometimes fails


lukehutch:
The first part of the test for issue 673 fails occasionally on Windows and Mac OS X, due to a race condition somewhere. When this test fails, the expected order is `b.zip, c.zip, a.zip` and the actual order is `a.zip, b.zip, c.zip`.

lukehutch:
(Comment added by Claude)

Fixed, and now covered by a deterministic test.

The cause: a `ClasspathElement` is created once per unique classpath entry via a singleton map, so anything recorded in its constructor came from whichever work unit won the race to create it. When `b.zip` was both a toplevel classpath entry and a `Class-Path` manifest entry of `a.zip`, the recorded classpath ordering key depended on the race, which is why the issue673 test failed intermittently on Windows and macOS.

`b1639a64` moved the ordering key out of the constructor into `ClasspathElement#addReference()`, so it is merged in from every reference: a toplevel reference always beats a reference from a parent classpath element, and between references of the same kind the lowest index wins.

Two follow-ups since:

- The classloader was still being set in the constructor, i.e. still decided by the race. That made the same directory reachable through two classloaders record a nondeterministic one, and `ClassInfo#loadClass()` intermittently loaded through the wrong one (this was behind an intermittent failure of the issue267 parent-last classloader test on macOS). Fixed in `060b79de` — the classloader now follows the winning reference too.
- `Issue810Test` only repeated the scan, and since the race cannot be forced it passed even with the fix reverted. `ClasspathElementReferenceTest` now tests the precedence rule directly; each half fails if the corresponding half of the fix is reverted.

PR #889, opened 2024-10-20
Fixes #888: ClasspathElementZip.getClasspathElementURI() returns inco…

### LINKED ISSUE #888: ClasspathElementZip.getClasspathElementURI() returns incorrect URIs under Windows
Version 4.8.177

[verbose.log](https://github.com/user-attachments/files/17447938/verbose.log)

When I invoke `ClassgraphElementZip.getClasspathElementURI()` on a JAR file on Windows, I get back: `file:/C:/Users/Gili/Documents/licensed-app/server/target/lib/client.browser-1.0-SNAPSHOT.jar!/index.html`

When I pass this URI to Jetty, it normalizes it to `file:///C:/Users/Gili/Documents/licensed-app/server/target/lib/client.browser-1.0-SNAPSHOT.jar!/` and complains `Base Resource should not be an alias`.

Upon further investigation, it looks like `file:/` is not actually a valid URI format. Is there any reason that ClassGraph is returning `file:/` instead of `file:///`? Is this by design? Or is this a bug? Thank you in advance.

cowwoc:
…rrect URIs under Windows.

* Use file:/// for absolute paths and file: for relative paths.
* Added "Test" suffix to unit tests that were being skipped by Surefire.


cowwoc:
Any idea why the build is failing on MacOS?

lukehutch:
I created a bug to track the macOS build failure, #890. It's probably a failure to find the Mac version of the Narcissus library (this is not related to your changes, it started happening recently, and I haven't had a chance to look into it yet).

The single windows failure is an example of #810, which again you can ignore (it is going to be tricky to track that one down, since it involves a race condition somewhere, which should be impossible given how ClassGraph was built, but apparently is not...).

src/test/java/io/github/classgraph/issues/issue305/Issue305Test.java
```diff
@@ -0,0 +1,90 @@
...
+import io.github.classgraph.ScanResult;
+
+/**
+ * Issue305.
+ */
+public class Issue305Test {
+    private ConsoleHandler errPrintStreamHandler = null;
+    private final Logger rootLogger = Logger.getLogger("");
+
+    /** Reset encapsulation circumvention method after each test. */
```
lukehutch, line 26: Comment is incorrect (came across via copy/paste from other test)

cowwoc: Hmm, for clarification, I don't think I added this comment. I think it was already there.

lukehutch:
Sorry to drop the ball, I thought I merged this already!

cowwoc:
No worries, thank you.

PR #893, opened 2024-11-08
[QuarkusClassLoaderHandler] Check for both `elements` and split element fields to accomdate Quarkus 3.15.x and earlier and 3.16.x and later.

michael-simons:
Fixes #891 (Sorry, typo and tagged the wrong issue)

PR #894, opened 2024-11-11
Polish `QuarkusClassLoaderHandler` and `FastPathResolver`.

michael-simons:
This is a follow up on issue #891 and the preceding PR #893 and brings two changes:

* The `QuarkusClassLoaderHandler` is a bit more explicit now about how the elements returned from the class loader are treated
* The `FastPathResolver` can be simplified using only one pattern

The latter was discovered while trying to get the custom URL scheme `quarkus:` that the Quarkus class loader uses for in-memory classes to work.

Ultimately, the work on using those in-memory classes, that are usually transformed classes such as session proxies, has been abandoned due to the fact that the current ClassGraph infrastructure can basically deal only with Jar and related entries, or directories.
To make the Quarkus in-memory classes work one would need to map them to ultimately into a Jar file and use an additional custom URL scheme, to "download" them again.
The alternative is additional implementations of `ClasspathElement`, but that would require a change on public API and this not only quite a bit out of my comfort zone, but also in all probability, not worth the effort.

I am happy to pick it if there's ever be a need, though.

The changes I had in addition (but those are only barely minimum, they won't work, as the scanner thinks that the urls are paths and does not use `openStream` or similar):

```java
    private static void findClasspathOrderForQuarkusClassloader(final ClassLoader classLoader,
                                                                final ClasspathOrder classpathOrder, final ScanSpec scanSpec, final LogNode log) {

        Collection<Object> elements = findQuarkusClassLoaderElements(classLoader, classpathOrder);

        for (final Object element : elements) {
            final String elementClassName = element.getClass().getName();
            final String fieldName = PRE_311_RESOURCE_BASED_ELEMENTS.get(elementClassName);
            if (fieldName != null) {
                classpathOrder.addClasspathEntry(classpathOrder.reflectionUtils.getFieldVal(false, element, fieldName),
                        classLoader, scanSpec, log);
            } else if("io.quarkus.bootstrap.classloading.MemoryClassPathElement".equals(elementClassName)) {
               handleInMemoryClassPathElements(classLoader, classpathOrder, scanSpec, log, element);
            } else {
                final Object rootPath = classpathOrder.reflectionUtils.invokeMethod(false, element, "getRoot");
                if (rootPath instanceof Path) {
                    classpathOrder.addClasspathEntry(rootPath, classLoader, scanSpec, log);
                }
            }
        }
    }

    @SuppressWarnings("unchecked")
    private static void handleInMemoryClassPathElements(final ClassLoader classLoader,
                                                        final ClasspathOrder classpathOrder, final ScanSpec scanSpec, final LogNode log, Object element) {
        // Those will be transformed classes, such as session proxies etc.
        Collection<String> resourceNames = (Collection<String>) classpathOrder.reflectionUtils.invokeMethod(false, element, "getProvidedResources");
        if (resourceNames == null) {
            return;
        }
        for (String resourceName : resourceNames) {
            Object resource = classpathOrder.reflectionUtils.invokeMethod(false, element, "getResource", String.class, resourceName);
            if (resource == null) {
                continue;
            }
            URL url = (URL) classpathOrder.reflectionUtils.invokeMethod(false, resource, "getUrl");
            classpathOrder.addClasspathEntryObject(url, classLoader, scanSpec, log);
        }
    }
```

### DIFF
--- a/src/main/java/nonapi/io/github/classgraph/classloaderhandler/QuarkusClassLoaderHandler.java
+++ b/src/main/java/nonapi/io/github/classgraph/classloaderhandler/QuarkusClassLoaderHandler.java
@@ -33,6 +33,8 @@ import java.net.URI;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 
 import nonapi.io.github.classgraph.classpath.ClassLoaderOrder;
@@ -52,6 +54,15 @@ class QuarkusClassLoaderHandler implements ClassLoaderHandler {
 
     // Classloader since Quarkus 1.13
     private static final String RUNNER_CLASSLOADER = "io.quarkus.bootstrap.runner.RunnerClassLoader";
+
+    // Class path elements prior to Quarkus 3.11
+    private static final Map<String, String> PRE_311_RESOURCE_BASED_ELEMENTS;
+    static {
+        Map<String, String> hlp = new HashMap<>();
+        hlp.put("io.quarkus.bootstrap.classloading.JarClassPathElement", "file");
+        hlp.put("io.quarkus.bootstrap.classloading.DirectoryClassPathElement", "root");
+        PRE_311_RESOURCE_BASED_ELEMENTS = Collections.unmodifiableMap(hlp);
+    }
 
     /**
      * Class cannot be constructed.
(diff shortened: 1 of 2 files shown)


michael-simons:
Of course I forget my main point: I would be quite happy if my previous change and potentially this additional change could go out as a release. Thank you! 🙇

lukehutch:
@michael-simons LGTM, I will push out a release later today -- sorry that I was offline for a few days! (I'll try to quickly get a Narcissus release before then, so ClassGraph can depend on the new version.)

michael-simons:
Nice, thank you. I set up the Narcissus pipelines and opened a PR there.

PR #900, opened 2025-01-24
Replace "String::replaceAll" with pre compiled pattern

kamilkrzywanski:
Replace the usage of String::replaceAll with a precompiled Pattern in the Jar utilities. The replaceAll method is invoked multiple times during the startup of my Spring Boot application, which impacts performance.


kamilkrzywanski:
@lukehutch How about this pr?

lukehutch:
LGTM, thanks!

PR #905, opened 2025-05-21
Update comment to specify that bootstrap class loader is used instead of context classloader

sebastiantoh:
The previous comment is inaccurate as the context classloader may have been modified. Instead, the comment should specify that bootstrap class loader is used

> If the parameter loader is null, the class is loaded through the bootstrap class loader

https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#forName(java.lang.String,%20boolean,%20java.lang.ClassLoader)


lukehutch:
Thanks!

PR #910, opened 2025-07-16
Revert breaking formatting changes of `pom.xml`

### LINKED ISSUE #909: The 4.8.181 release on Maven Central no longer provides an OSGi MANIFEST.MF
Looking at the latest release

https://repo1.maven.org/maven2/io/github/classgraph/classgraph/4.8.181/classgraph-4.8.181.jar

the MANIFEST.MF looks like this:

Image

Looking at the previous release

https://repo1.maven.org/maven2/io/github/classgraph/classgraph/4.8.180/classgraph-4.8.180.jar

it has a well-formed OSGi MANIFEST.MF:

Image

I looked in the commit history, but I don't really see what's changed that causes this difference.  But I assume this is accidental and not intentional.

We've been consuming classgraph in Orbit:

https://github.com/eclipse-orbit/orbit-simrel/blob/9745ac69e7f7cbfb40b9a179fe5b876a0f871f6f/maven-osgi/tp/Maven.target#L443-L448

When I updated this automatically to the new version the build fails because it specifies `missingManifest="error"`:

https://github.com/eclipse-orbit/orbit-simrel/blob/9745ac69e7f7cbfb40b9a179fe5b876a0f871f6f/maven-osgi/tp/Maven.target#L5

The classgraph library is used by https://github.com/eclipse-xtext/xtext:

FYI  @LorenzoBettini @cdietrich

Please let me know what to expect in terms for future version.

freya022:
Cancels what seems to be 80-char hard wraps, which caused some strings to no longer be the same, breaking generation of manifests.

```
[INFO] --- antrun:3.1.0:run (add-module-info-to-jar) @ classgraph ---
[INFO] Executing tasks
[ERROR]       [jar] Manifest is invalid: Manifest line "                for JVM languages." is not valid as it does not contain a name and a value separated by ': '
```

Fixes #909

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,6 +1,4 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 
 	<groupId>io.github.classgraph</groupId>
@@ -8,8 +6,7 @@
 	<version>4.8.181</version>
 	<name>ClassGraph</name>
 
-	<description>The uber-fast, ultra-lightweight classpath and module scanner
-		for JVM languages.</description>
+	<description>The uber-fast, ultra-lightweight classpath and module scanner for JVM languages.</description>
 
 	<url>https://github.com/classgraph/classgraph</url>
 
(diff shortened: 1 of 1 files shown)


michael-simons:
Oh cool, I just wanted to created a PR for this. 

thank you @freya022 

However, this would be enough to fix:

```
diff --git a/pom.xml b/pom.xml
index b8331e3b..8554314c 100644
--- a/pom.xml
+++ b/pom.xml
@@ -8,8 +8,7 @@
 	<version>4.8.181</version>
 	<name>ClassGraph</name>
 
-	<description>The uber-fast, ultra-lightweight classpath and module scanner
-		for JVM languages.</description>
+	<description>The uber-fast, ultra-lightweight classpath and module scanner for JVM languages.</description>
 
 	<url>https://github.com/classgraph/classgraph</url>

```

Please take note @lukehutch release 4.8.181 is broken (for us it fails on the module path, without a valid manifest, the Multi-Release Jar is invalid, and while the module-info is still present, JVM won't care…)

freya022:
I know this would be enough to fix it, but it's likely that those formatting changes were not intended, an arbitrary 80-char hard wrap makes a lot of these comments hard to read.
The break points do not make sense and so were likely automatically created, so I reverted all formatting changes, keeping only the maven publishing changes.

michael-simons:
Makes sense. And I agree wrt line breaks in a Maven document.

freya022:
It would also be nice if `maven-antrun-plugin` failed on errors, i've looked at the [JAR task](https://ant.apache.org/manual/Tasks/jar.html) docs but there isn't anything to make it fail on warnings/errors. The `failOnError` of the plugin itself doesn't make it fail either.

lukehutch:
Sorry about that -- I switched from Eclipse to VS Code, and didn't bring over the formatting settings.

I'm very surprised that maven-jar-plugin does not strip newlines out of text that is copied into a manifest file -- I'll try to file an issue on that.

PR #919, opened 2025-09-23
feat(scan): propagate ScanResult to `ModuleInfo` and `PackageInfo` annotations

aurimasniekis:
Introduce `setScanResult` methods in `ModuleInfo` and `PackageInfo` to assign `ScanResult` context to their associated annotations. This enables annotations to access scanning metadata when class info is enabled. Called from `ScanResult` during post-processing.


lukehutch:
Thanks, good catch! This looks right to me.

Issue #920, opened 2025-09-23
ClassGraph returns superclass constructors when asked for subclass constructors even if the subclass has a non-public constructor with the same signature as a super's constructor


chubert-ch:
Already got a fix, just making this issue to have an issue number:
```
    @Test
    void test() {
        MethodInfoList constructors = new ClassGraph()
                .enableAnnotationInfo()
                .enableSystemJarsAndModules()
                .enableClassInfo()
                .enableMethodInfo()
                .scan()
                .getClassInfo("java.io.ObjectOutputStream")
                .getConstructorInfo();
        for (MethodInfo constructor : constructors) {
            if (constructor.getParameterInfo().length == 0) {
                // The no args constructor of ObjectOutputStream is protected
                assertEquals(Modifier.PROTECTED, constructor.getModifiers(), "The no-args constructor of ObjectOutputStream should read as `protected`");
            }
        }
    }
```

Basically, the constructor info returned is for the superclass constructor, but given constructors aren't inherited this seems more misleading than useful.

Also worth noting that the javadoc for `getMethodInfo(final String methodName)` says:

> Returns information on the method(s) or constructor(s) of the given name declared by this class, but not by its interfaces or superclasses.

but currently it will in fact return a public superclass constructor if the subclass one is lower visibility and `.ignoreMethodVisibility()` isn't set.

lukehutch:
Thanks -- merged! I think this also fixes your "Also worth noting" part (i.e. the doc was correct but the behavior was wrong before your change)?

lukehutch:
Released in 4.8.185 (sorry for the long delay!).

PR #921, opened 2025-09-23
[Issue 920] Updating getMethodInfo to not get superconstructors

chubert-ch:
ClassGraph returns the MethodInfo of some (but not all!) superclass constructors when asked for a Class's constructors.

Particularly confusing for users if there's a non-public constructor with the same signature as a constructor in the superclass and `.ignoreMethodVisibility()` isn't set. This is the only case where an overridden method is returned from `getMethodInfo()` and is because constructors are the only case where a method can have the same signature as a superclass method, but have lower visibility.

[Issue link](https://github.com/classgraph/classgraph/issues/920)


lukehutch:
Good catch, thanks for the fix!

Issue #906, opened 2025-06-05
Ability to handle custom ClassLoader(s)


freya022:
Hi, as I am making my own "restart" `ClassLoader`, which is basically the same as Spring Boot's `RestartClassLoader`, I have found the need to prioritize my class loader, similar to what `SpringBootRestartClassLoaderHandler` does.

However this can only be done by contributing here, and if I do that for my small library, others are also able to, and I think at this point it becomes kind of unmanageable, in addition to having to wait for it to be accepted and released.

Basically my question is: Should I still PR it, or should there be a better way to do this, such as letting the user add their own `ClassLoaderHandler` before scanning?

lukehutch:
Hi, is it possible to subclass `URClassLoader` in your case? If so, this should work out of the box.

However, correct, there is currently no way to register your own `ClassLoaderHandler`, and it would actually be awesome to be able to do so. 

Would you be interested in working on a PR to add that functionality?

freya022:
I do use a `URLClassLoader` subclass, the issue is that it is *last* in ClassGraph's `ClassLoader` order.

![Image](https://github.com/user-attachments/assets/6009a19b-b1fa-44fd-870b-01f5feb17668)

This causes the classes to always be loaded using the `AppClassLoader`, and so my classes are not refreshed on hot restarts.

### Alternatives

The goal is to let third parties easily add their `ClassLoaderHandler`, not to redefine how the `ClassLoader`s are handled.

It think it should load  `ClassLoaderHandler` using `ServiceLoader` , though I'm not sure why 15086253034434f8ab5d6b1d1fc62eff70b59558 changed from interface methods to reflectively-accessed static methods.

The `ClassLoaderHandler` would use interface methods (again), but also define a priority as well as getters to populate the `AUTOMATIC_LIB_DIR_PREFIXES` and `AUTOMATIC_PACKAGE_ROOT_PREFIXES` fields

The "priority" would be a `int`, with the existing handler order being defined as:
- `1000`: Handlers specific to third party integrations (Ant, JBoss, Spring, Tomcat...), all of them will share this priority as their order between them doesn't matter
- `700` (test only): `ParentLastDelegationOrderTestClassLoaderHandler`
- `600`: JPMS support
- `500`: `URLClassLoader` support
- `0`: `ClassGraphClassLoader`

Feel free to submit your own idea.

lukehutch:
[...]

Issue #929, opened 2025-12-24
It is not possible to find classes in the classes package within the classes directory using Maven.


mcnew:
How to reproduce:

1. Create the `interface test.Alpha`
2. Create the `class classes.AlphaImpl implements test.Alpha`.

The following code cannot find the class classes.AlphaImpl

```
		try (ScanResult scanResult = new ClassGraph()
				.verbose()
				.enableClassInfo() // Scan classes
				.acceptPackages("classes")
				.scan()) {
			for (ClassInfo routeClassInfo : scanResult.getClassesImplementing(test.Alpha.class)) {
				System.out.println(routeClassInfo.getName());
			}
		}
```

If the package for the AlphaImpl class is changed to other (`other.AlphaImpl`) and the code is updated to search in the `other` package, the class is found.

The issue appears to be related to this line [ClasspathElementDir.java:351](https://github.com/classgraph/classgraph/blob/3ed377e7f8455fd85eac3f4359f3aaa1433f2a28/src/main/java/io/github/classgraph/ClasspathElementDir.java#L351), where `classes/` already exists in `nestedClasspathRootPrefixes`

lukehutch:
(Comment added by Claude)

Confirmed and fixed in d08060ac.

Your diagnosis was right: `classes/` is stripped as an Ant-style package root, so `classes.AlphaImpl` was scanned as if it were the top-level class `AlphaImpl`. But `classes` and `test-classes` are legal Java package names, so directory structure alone can't tell a package root from a package with the same name.

The fix reads the class name declared by the first classfile beneath the candidate package root: if the declared name still starts with the candidate's name, it's a package, not a package root, and the prefix is not stripped. This is applied to both directory and jar classpath elements. Ant-style layouts (a real `classes/` package root containing `com/xyz/Beta.class`) still work.

PR #934, opened 2026-06-12
`ClassLoaderHandler`: Use interface instead of reflection

freya022:
We've discussed it a bit in here https://github.com/classgraph/classgraph/issues/906#issuecomment-3027438679

It's not really a followup to the issue, I just figured I'd PR improvements like these as i've been looking to optimize my reflection code.

I haven't bothered using `ServiceLoader`, doesn't seem necessary right now.

### DIFF
--- a/src/main/java/nonapi/io/github/classgraph/classloaderhandler/AntClassLoaderHandler.java
+++ b/src/main/java/nonapi/io/github/classgraph/classloaderhandler/AntClassLoaderHandler.java
@@ -36,54 +36,21 @@ import nonapi.io.github.classgraph.utils.LogNode;
 
 /** Extract classpath entries from the Ant ClassLoader. */
 class AntClassLoaderHandler implements ClassLoaderHandler {
-    /** Class cannot be constructed. */
-    private AntClassLoaderHandler() {
-    }
 
-    /**
-     * Check whether this {@link ClassLoaderHandler} can handle a given {@link ClassLoader}.
-     *
-     * @param classLoaderClass
-     *            the {@link ClassLoader} class or one of its superclasses.
-     * @param log
-     *            the log
-     * @return true if this {@link ClassLoaderHandler} can handle the {@link ClassLoader}.
-     */
-    public static boolean canHandle(final Class<?> classLoaderClass, final LogNode log) {
+    @Override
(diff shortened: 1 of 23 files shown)


lukehutch:
(Comment added by Claude)

Thanks for this -- the refactor itself is sound and I would like to take it. A few things before it can be merged.

**Public API narrowing.** The PR makes `ClassLoaderHandler` package-private and removes the `classLoaderHandlerClass` field from `ClassLoaderHandlerRegistryEntry`. Neither `module-info.java` nor the OSGi `Export-Package` header exports that package, so module-path and OSGi users are unaffected -- but plain-classpath users who implemented a custom `ClassLoaderHandler` would break at compile time. I am fine with that given the package is not exported, but it should be called out in the release notes, so please leave a note in the PR description saying so.

**Requests:**
- Please rebase on current `latest` -- there have been several commits since this branched.
- Mark the new method parameters `final`, to match the surrounding code.
- Add a Javadoc comment on `getHandlerName()`.

Once those are in I will merge it.

lukehutch:
(Comment added by Claude)

I did the rebase myself rather than asking you to -- this is now on `latest` as commit 08bd83c4, with your authorship preserved. Full test suite green (312 tests, 0 failures).

Three things came up in the rebase, all handled:

- `getPackageRootPrefixes()` was added to the handlers after your branch was cut (#929), still as a reflectively-invoked `static`. I made it a fourth interface method, in the spirit of the PR, so `ClassLoaderHandlerRegistryEntry` now reads it once at construction from the handler instance.
- Conflicts in `SpringBootRestartClassLoaderHandler` and `ClasspathFinder`, resolved to keep both sides: the package root prefixes are still set on `ClasspathOrder` around `findClasspathOrder()`, and the handler name in the log now comes from your `getHandlerName()`.
- The changes I had asked for: `final` parameters, wrapping at 120 columns to match the surrounding code, and Javadoc on `getHandlerName()` and `getPackageRootPrefixes()`.

The commit message notes the API narrowing (`ClassLoaderHandler` becomes package-private, `classLoaderHandlerClass` is replaced by `classLoaderHandler`), so it makes the release notes.

Nice cleanup -- it removes 880 lines. Thanks!

PR #935, opened 2026-06-12
Optimize and remember resolved/sanitized paths

freya022:
I've found that quite some time was spent on resolving/sanitizing paths, it did it 5 times for each .class, now this PR brings it to only once. `.jar` paths are still resolved multiple times but I didn't see an easy fix for it, and the impact seems to be minimal.

The changes in `FileUtils` only changes detecting when to sanitize, it should be equivalent to the previous code but much faster.

No clue why CI is failing, `mvn clean test` passes on my machine. Also the MacOS and Ubuntu CIs don't show the same errors lol

### DIFF
--- a/src/main/java/io/github/classgraph/ClasspathElementDir.java
+++ b/src/main/java/io/github/classgraph/ClasspathElementDir.java
@@ -65,6 +65,8 @@ import nonapi.io.github.classgraph.utils.VersionFinder;
 
 /** A directory classpath element, using the {@link Path} API. */
 class ClasspathElementDir extends ClasspathElement {
+    private static final int NOT_YET_LOADED_LENGTH = -2;
+
     /** The directory at the root of the classpath element. */
     private final Path classpathEltPath;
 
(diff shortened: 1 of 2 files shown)


GedMarc:
@freya022 One of the things I'd be concerned with is when the system is running under JRT:// and not JAR:// 
A lot went into properly identifying the module path when running as native runtimes - 

I imagine these are where the tests are failing looking at the changes

GedMarc:
This test may need updating if you added sorting to the jar scan (which may add read times, this library is optimized for speed)

Error:  Failures: 
Error:    Issue673Test.testResourcesCanBeRead:26 
expected: ["b.zip", "c.zip", "a.zip"]
 but was: ["a.zip", "b.zip", "c.zip"]
 
 Would be curious as to why they are being sorted now

freya022:
> One of the things I'd be concerned with is when the system is running under JRT:// and not JAR://
A lot went into properly identifying the module path when running as native runtimes

Not sure what your concern is, but this PR does not change any behavior, it only memoizes the result and optimizes finding out if a path needs to be sanitized (as most of them don't), but not how they are sanitized.

The fail at `Issue673Test.testResourcesCanBeRead:26` is also failing in #937, and both I and his PR didn't change anything related to that.

lukehutch:
(Comment added by Claude)

Thanks for this. The PR is in two halves, and they need to be split.

**The `ClasspathElementDir` half is sound.** Hoisting the repeated `subPathRelativeStr` computation is a straightforward win -- I checked the expression at lines 505-506 and it is identical to the one being replaced. Happy to take that on its own.

**The `sanitizeEntryPath` half is a regression.** Replacing the segment-scanning detector with

```java
path.contains("//") || path.contains("!!") || path.contains(".!") || path.contains("./")
```

misses trailing `.` and `..` segments, because those are not followed by a separator so none of the four substrings appear. Measured against the current code:

| input | current | with this PR |
|---|---|---|
| `a/b/..` | `a` | `a/b/..` (unsanitized) |
| `foo/.` | `foo` | `foo/.` (unsanitized) |
| `foo/..` | `` (empty) | `foo/..` (unsanitized) |

Those paths would then be compared and cached as-is, so two spellings of the same entry stop matching.

Could you resubmit the `ClasspathElementDir` change on its own? I would take that as-is.

lukehutch:
(Comment added by Claude)

I have gone ahead and done the split myself, so you do not have to -- the `ClasspathElementDir` half is now on `latest` as commit 322431b4, credited to you. I simplified the plumbing slightly (two overloads of `newResource` rather than a three-method chain) but the change is yours: pass the already-computed `subPathRelativeStr` through instead of recomputing the identical expression in `getPath()`.

The `sanitizeEntryPath` half is not included, for the reason in my previous comment. Closing this PR as partially merged.

Thanks!

lukehutch:
> This test may need updating if you added sorting to the jar scan (which may add read times, this library is optimized for speed)
> 
> Error: Failures: Error: Issue673Test.testResourcesCanBeRead:26 expected: ["b.zip", "c.zip", "a.zip"] but was: ["a.zip", "b.zip", "c.zip"]
> 
> Would be curious as to why they are being sorted now

@GedMarc that was fixed in #810.

PR #936, opened 2026-06-25
Lower-case URL scheme with Locale.ROOT to avoid Turkish-locale mismatch

koteshyelamati:
`ScanSpec.enableURLScheme(scheme)` stores `scheme.toLowerCase()` using the JVM default locale. Scheme matching later compares against `URI.getScheme()`, which is ASCII lower-case (e.g. `file`). Under a Turkish locale, `"FILE".toLowerCase()` produces `"fıle"` (dotless i), so `enableURLScheme("FILE")` no longer matches the `file` scheme and those classpath entries are silently skipped. Any scheme containing `I`/`i` is affected (the classic Turkish-locale bug).

Fix: lower-case with `Locale.ROOT` so the stored scheme is always ASCII.

Adds `ScanSpecLocaleTest`, which sets the default locale to tr-TR, calls `enableURLScheme("FILE")`, and asserts the stored set contains `file`. The test fails on `latest` and passes with this change (verified locally with Maven).

### DIFF
--- a/src/main/java/nonapi/io/github/classgraph/scanspec/ScanSpec.java
+++ b/src/main/java/nonapi/io/github/classgraph/scanspec/ScanSpec.java
@@ -38,6 +38,7 @@ import java.nio.channels.FileChannel;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Locale;
 import java.util.List;
 import java.util.Set;
 
@@ -357,7 +358,7 @@ public class ScanSpec {
         if (allowedURLSchemes == null) {
             allowedURLSchemes = new HashSet<>();
         }
-        allowedURLSchemes.add(scheme.toLowerCase());
+        allowedURLSchemes.add(scheme.toLowerCase(Locale.ROOT));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lukehutch:
(Comment added by Claude)

Thanks! Confirmed on JDK 8 under `tr_TR`: `"FILE".toLowerCase()` returns `"fıle"` (dotless ı), so a scheme registered that way could never match. Merged.

I had also thought there was a second half to this in `NestedJarHandler`, which compares `allowedURLSchemes.contains(scheme)` against the scheme taken from the URL without lower-casing it. On investigation that turns out not to be reachable, so there is nothing further to fix. Every path that reaches that check has already been normalised: `http`/`https` are lower-cased by `FastPathResolver.resolve()`, anything arriving as a `java.net.URL` is lower-cased by `URL` itself, and a scheme that neither of those recognises (`FTP://`, `ZZ://`) never survives classpath-element parsing with its case intact -- it is either split on the `:` as a path separator or dropped. The check is not dead code, though: scanning `ftp://host/x.jar` without `enableURLScheme("ftp")` does produce the expected "has not been enabled" error. It simply never sees anything but lower case.

PR #937, opened 2026-07-11
Fix annotation equals NPE after ScanResult is closed (#930)

### LINKED ISSUE #930: [bug] Can't use methods on annotations after scan result is closed.
I've found a bug with annotations that are loaded via classgraph.

When invoking the `equals` method on the annotation after the scan result is closed, a `NullPointerException` is thrown [here](https://github.com/classgraph/classgraph/blob/latest/src/main/java/io/github/classgraph/AnnotationInfo.java#L360).

From what I can tell, the reason for this is that the [preceding code](https://github.com/classgraph/classgraph/blob/latest/src/main/java/io/github/classgraph/AnnotationInfo.java#L354-L356) is checking the wrong thing. When the scan result is closed, `annotationInfo.scanResult` is not touched. However, `annotationInfo.scanResult.reflectionUtils` IS set to null. I think the fix is to change the condition of that ternary to check for `annotationInfo.scanResult.reflectionUtils == null`.

I've created a small reproducible example here: https://github.com/Skater901/classgraph-annotation-equals-problem/blob/main/src/main/kotlin/au/com/skater901/main.kt

kamilkrzywanski:
ScanResult.close() nulls reflectionUtils but leaves AnnotationInfo.scanResult non-null, so AnnotationInvocationHandler.equals used a null ReflectionUtils. Fall back to a new ReflectionUtils when reflectionUtils has been cleared. Add Issue930Test covering equals after close.
Fixes: #930

### DIFF
--- a/src/main/java/io/github/classgraph/AnnotationInfo.java
+++ b/src/main/java/io/github/classgraph/AnnotationInfo.java
@@ -351,9 +351,12 @@ public class AnnotationInfo extends ScanResultObject implements Comparable<Annot
                     } else if (!annotationClass.isInstance(args[0])) {
                         return false;
                     }
+                    // ScanResult#close() nulls reflectionUtils but leaves scanResult non-null on
+                    // AnnotationInfo, so also fall back when reflectionUtils has been cleared (#930).
                     final ReflectionUtils reflectionUtils = annotationInfo.scanResult == null
-                            ? new ReflectionUtils()
-                            : annotationInfo.scanResult.reflectionUtils;
+                            || annotationInfo.scanResult.reflectionUtils == null //
+                                    ? new ReflectionUtils()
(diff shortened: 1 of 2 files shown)


kamilkrzywanski:
Seems that CI failed on flay test (checked locally and 1/500 failed) can someone rerun CI?

lukehutch:
(Comment added by Claude)

Thanks for this -- the same problem was fixed in 4.8.186, as commit a00d2056. That fix is a superset of this PR: as well as the `AnnotationInfo` equals handler you patched here, it routes the lookup through a new `ScanResult.getReflectionUtils(ScanResult)` helper so the field is read once rather than at each call site, and it covers `ClassInfo#getEnumConstantObjects`, which had the same post-close NPE.

Closing as already fixed. Thanks for the report and the test case.

PR #938, opened 2026-07-11
Fix ThreadGroup memory leak in default SimpleThreadFactory (#931)

### LINKED ISSUE #931: ThreadGroup and Thread memory leak when using the default SimpleThreadFactory
Hi there!
While using this amazing library during a mass-failure event, we noticed how our service (which is continuously scanning the classpath using a `new ClassGraph()....scan()` invocation, to find file changes, was displaying a memory leak.

Upon looking at the histogram, and after being able to reproduce the issue in a controlled environment where I was able to get a heap dump, we noticed that the biggest offenders were the following (this was after a few days):

```
 num     #instances         #bytes  class name (module)
-------------------------------------------------------
   1:       8973221      430714608  java.lang.ThreadGroup (java.base@17.0.15.0.101)
   2:       8973220      287151232  [Ljava.lang.Thread; (java.base@17.0.15.0.101)
```

These thread groups, as it came out from the heap dump, were all called `ClassGraph-thread-group` and came from the [newThread()](https://github.com/classgraph/classgraph/blob/3ed377e7f8455fd85eac3f4359f3aaa1433f2a28/src/main/java/nonapi/io/github/classgraph/concurrency/SimpleThreadFactory.java#L70-L89) method in `SimpleThreadFactory`

Those thread and groups seem to have a GC root of `JNI Global` which is preventing them from being GC'd.

The easy fix was to use an ad-hoc fixed thread pool, but I was wondering if I missed something in the docs or in the examples that would have prevented this issue.

Thank you!

kamilkrzywanski:
## Summary

`SimpleThreadFactory.newThread()` created a new `ThreadGroup("ClassGraph-thread-group")` for every worker thread when no `SecurityManager` was present. Default `ClassGraph.scan()` builds a fresh `AutoCloseableExecutorService` each call, so long-running apps that scan repeatedly accumulated millions of `ThreadGroup` / `Thread[]` instances that were not GC'd (JNI Global roots).

Match `Executors.defaultThreadFactory()`: use the security manager's thread group when present, otherwise the current thread's group. Do not allocate a fresh `ThreadGroup` per thread.

## Test plan

- [x] Added `Issue931Test` — 20 default scans, asserts no `ClassGraph-thread-group` instances are created
- [x] `./mvnw -Dtest=Issue931Test test` passes locally

Fixes: #931

### DIFF
--- a/src/main/java/nonapi/io/github/classgraph/concurrency/SimpleThreadFactory.java
+++ b/src/main/java/nonapi/io/github/classgraph/concurrency/SimpleThreadFactory.java
@@ -62,6 +62,12 @@ public class SimpleThreadFactory implements java.util.concurrent.ThreadFactory {
     /**
      * New thread.
      *
+     * <p>
+     * Matches the {@link java.util.concurrent.Executors#defaultThreadFactory()} pattern: use the
+     * {@link SecurityManager}'s thread group when present, otherwise the current thread's group. Do
+     * <em>not</em> allocate a fresh {@link ThreadGroup} per thread — that leaks {@code ThreadGroup}
+     * (and related) instances under repeated scans with the default executor (#931).
+     *
      * @param runnable
      *            the runnable
      * @return the thread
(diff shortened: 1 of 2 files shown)


lukehutch:
(Comment added by Claude)

Thanks for this -- the same fix shipped in 4.8.186, as commit 847a53f4. It makes the same change you did here: use `Thread.currentThread().getThreadGroup()` as the fallback instead of allocating a `new ThreadGroup("ClassGraph-thread-group")` per thread, keeping the SecurityManager group when there is one. So the leak is already gone as of that release.

The one thing your PR has that the release does not is the end-to-end test -- the shipped fix is covered by a unit test on `SimpleThreadFactory` rather than by counting live thread groups after repeated scans. If you would like to resubmit `Issue931Test` on its own, I would take it.

Closing, since the fix itself is already in. Thanks for tracking this down and for the heap-dump detail in the issue.

